PR #47, opened 2014-09-17
Adds exception throwing method support

Raphfrk:
This would allow us to automatically set any un-implemented methods to throw an UnsupportedOperationException() by scanning for any missing methods in the interface.


Zidane:
Closing this as we don't want the impl being tested with API methods unimplemented.

PR #88, opened 2014-12-22
Add getUniqueId to MixinEntity.

kramerc:
Implements `getUniqueId()` for players. I'm assuming that `getGameProfile().getId()` is the best way to get this unique identifier based on the information at http://www.minecraftforge.net/forum/index.php?topic=22491.0

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/mod/mixin/entity/MixinEntity.java
@@ -25,6 +25,8 @@
 package org.spongepowered.mod.mixin.entity;
 
 import java.util.ArrayDeque;
+import java.util.UUID;
+
 import javax.annotation.Nullable;
 
 import net.minecraft.entity.player.EntityPlayerMP;
@@ -88,6 +90,13 @@ public abstract class MixinEntity implements Entity, ISpongeEntity {
     protected abstract void shadow$setRotation(float yaw, float pitch);
     @Shadow
     public abstract void mountEntity(net.minecraft.entity.Entity entityIn);
+    @Shadow
+    public abstract UUID getUniqueID();
+
+    @Override
+    public UUID getUniqueId() {
+        return this.getUniqueID();
+    }
 
     @Override
     public World getWorld() {


simon816:
This is incorrect
`getUniqueId` is defined by `Entity` (it implements `Identifiable`) as the UUID of the in-game entity.

The `GameProfile` UUID should intead be used in the `User` class for the user's persistent UUID.

However, I've spotted a problem where the same method `getUniqueId` exists in `User` and `Entity` and therefore `Player` cannot extend both

kenzierocks:
`getUserUUID`?

simon816:
On further investigation it seems the in-game UUID and the persistent User UUID are the same.

But this PR should implement `getUniqueId` in `MixinEntity` regardless

ryantheleach:
public UUID getUniqueId() { User.super.getUniqueId(); }

Or does that only work in java 1.8?

kramerc:
@simon816 Thanks, I've updated the PR to implement `getUniqueId` in `MixinEntity` instead.

src/main/java/org/spongepowered/mod/mixin/entity/MixinEntity.java
```diff
@@ -90,6 +92,13 @@
     public abstract void mountEntity(net.minecraft.entity.Entity entityIn);
 
     @Override
+    public UUID getUniqueId() {
```
Mumfrey: You cannot ```@Override``` this method to implement the method from the interface because after obfuscation it will get the MCP name. Please consult [the mixin examples](https://github.com/SpongePowered/Sponge/blob/master/src/example/java/org/spongepowered/mod/mixin/MixinEntityPlayerExample.java) for how to deal with conflicting methods with the **same** signature/name.

src/main/java/org/spongepowered/mod/mixin/entity/MixinEntity.java
```diff
@@ -90,6 +92,13 @@
...
+        Entity spongeEntity = (Entity)this;
```
Mumfrey: What is all this weird casting? Why don't you just return ```this.getUniqueID()``` ?

kramerc: The `getUniqueID` method is in NMS. Using `this.getUniqueID()` isn't possible because it's not implemented in this class or any superclass it inherits. The weird casting is based on this convention further down: https://github.com/SpongePowered/Sponge/blob/52d67d0b757950b09d489415928eaa8b002944c4/src/main/java/org/spongepowered/mod/mixin/entity/MixinEntity.java#L108-109

Minecrell:
I've implemented this together with #143 and totally forgot about this pull request, so this was fixed as of https://github.com/SpongePowered/Sponge/commit/c21d44c390da4412c23b9bf250578ada3ef6f504. Thank you for your contribution!

PR #100, opened 2014-12-29
Implemented the extended particles api.

Cybermaxke:
I implemented the basics of the particle api, only the Viewer should be implemented in order to send the packets created by the SpongeParticleHelper.

### DIFF
--- a/src/main/java/org/spongepowered/mod/effect/particle/SpongeParticleHelper.java
+++ b/src/main/java/org/spongepowered/mod/effect/particle/SpongeParticleHelper.java
@@ -39,6 +39,7 @@ import org.spongepowered.api.item.ItemBlock;
 import org.spongepowered.api.item.ItemType;
 import org.spongepowered.api.item.inventory.ItemStack;
 
+import com.flowpowered.math.vector.Vector3d;
 import com.flowpowered.math.vector.Vector3f;
 import com.google.common.collect.Lists;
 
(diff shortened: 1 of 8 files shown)


src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -266,23 +274,30 @@ public TradeOfferBuilder getTradeOfferBuilder() {
...
-        //TODO implement.
-        return null;
+        return new ArrayList<ParticleType>(this.particleByName.values());
     }
 
     @Override
     public ParticleEffectBuilder getParticleEffectBuilder(ParticleType particle) {
-        // TODO
-        return null;
+        Preconditions.checkNotNull(particle);
```
simon816: Should be statically imported

src/main/java/org/spongepowered/mod/effect/particle/SpongeParticleHelper.java
```diff
@@ -0,0 +1,196 @@
...
+    /**
+     * Gets the list of packets that are needed to spawn the particle effect at the position.
+     * This method should tries to minimize the amount of packets for a better performance and
+     * lower bandwidth use.
+     * 
+     * @param effect The particle effect
+     * @param position The position
+     * @return The packets
+     */
+    public static List<Packet> toPackets(SpongeParticleEffect effect, Vector3d position) {
```
simon816: Is it necessary to create the packets in Sponge?
Most of this can be done using `net.minecraft.world.World#spawnParticle` (`WorldServer#func_180505_a` for the server impl)
Perhaps a tweak to the world class would be better than this complex packet creation

Cybermaxke: @simon816 We would be able to do this for the server implementation, but WorldServer#func_180505_a creates the packets differently which will break the colors/sizes/notes if the count is greater then zero. It is also possible to call the WorldServer#func_180505_a for the count of particles if needed.

bloodmc:
Merged thanks!

PR #104, opened 2014-12-31
Implement Weather

AIOOB:
All methods added by WeatherVolume have been tested and work. I have also accidentally formatted MixinWorld, say if I should change it back.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
+++ b/src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
@@ -24,9 +24,10 @@
  */
 package org.spongepowered.mod.mixin.world;
 
-import com.flowpowered.math.vector.Vector3d;
-import com.flowpowered.math.vector.Vector3i;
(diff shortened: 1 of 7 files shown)


src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
```diff
@@ -110,4 +115,73 @@ public Biome getBiome(Vector3d position) {
...
+    }
+
+    @Override
+    public long getRunningDuration() {
+        return this.worldInfo.getWorldTotalTime() - this.weatherStartTime;
+    }
+
+    @Override
+    public void forecast(Weather weather) {
+        this.forecast(weather, (300 + (new Random()).nextInt(600)) * 20);
```
ryantheleach: Is there anyway to hook into that code? Otherwise leaving a comment referencing where to find the code in 1.8 may help with future updates.

bloodmc:
Merged thanks.

PR #106, opened 2015-01-01
Implement selectors

AIOOB:


src/main/java/org/spongepowered/mod/text/selector/SpongeSelector.java
```diff
@@ -0,0 +1,92 @@
...
+
+    @Override
+    public Map<String, String> getArguments() {
+        return this.arguments;
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public List<Entity> resolve(Location location) throws IllegalArgumentException {
+        Preconditions.checkArgument(this.requiresLocation == true && location == null);
```
kenzierocks: Should be `!this.requiresLocation || location != null`.

Also, Preconditions should be statically imported.

src/main/java/org/spongepowered/mod/text/selector/SpongeSelector.java
```diff
@@ -0,0 +1,92 @@
...
+    @Override
+    public boolean requiresLocation() {
+        return this.requiresLocation;
+    }
+
+    @Override
+    public String asString() {
+        StringBuilder sb = new StringBuilder();
+        for (Entry<String, String> entry : this.arguments.entrySet()) {
+            sb.append(entry.getKey()).append("=").append(entry.getValue());
```
kenzierocks: This will not insert commas as required.

src/main/java/org/spongepowered/api/text/selector/SpongeSelectorFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+import java.util.regex.Pattern;
+
+import org.spongepowered.api.util.annotation.NonnullByDefault;
+import org.spongepowered.mod.text.selector.SpongeSelectorBuilder;
+
+@NonnullByDefault
+public class SpongeSelectorFactory implements SelectorFactory {
+
+    private static final Pattern tokenPattern = Pattern.compile("^@([pare])(?:\\[([\\w=,!-]*)\\])?$");
```
kenzierocks: Doesn't match decimals, which are allowed.

src/main/java/org/spongepowered/api/text/selector/SpongeSelectorFactory.java
```diff
@@ -0,0 +1,56 @@
...
+
+    @Override
+    public SelectorBuilder createEmptyBuilder() {
+        return new SpongeSelectorBuilder();
+    }
+
+    @Override
+    public Selector parseRawSelector(String selector) {
+        if (tokenPattern.matcher(selector).matches()) {
+            SelectorBuilder builder = createEmptyBuilder().type(SelectorTypes.valueOf(selector.substring(1, 2)).get());
```
kenzierocks: Doesn't account for this odd behavior:
>If the first few arguments (up to four) are specified simply as values without identifying the arguments, they are assumed to be, in order, x, y, z, and r. Thus, the following two commands are identical:
/gamemode creative @a[x=10,y=20,z=30,r=4]
/gamemode creative @a[10,20,30,4]

src/main/java/org/spongepowered/api/text/selector/SpongeSelectorFactory.java
```diff
@@ -0,0 +1,56 @@
...
+            String argString = selector.substring(selector.indexOf('[') + 1, selector.lastIndexOf(']'));
+            for (String argument : argString.split(",")) {
+                int index = argument.indexOf('=');
+                builder.addArgument(argument.substring(0, index), argument.substring(index) + 1);
+            }
+            return builder.build();
+        }
+        return null;
```
kenzierocks: Perhaps we need an exception, because this can't be null.

src/main/java/org/spongepowered/mod/text/selector/FakeCommandSender.java
```diff
@@ -0,0 +1,99 @@
...
+import net.minecraft.util.BlockPos;
+import net.minecraft.util.IChatComponent;
+import net.minecraft.util.Vec3;
+import net.minecraft.world.World;
+
+import org.spongepowered.api.world.Location;
+
+import com.flowpowered.math.vector.Vector3d;
+
+class FakeCommandSender implements ICommandSender {
```
kenzierocks: Do we need this for this PR?

src/main/java/org/spongepowered/mod/text/selector/SpongeSelectorBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public SelectorBuilder type(EntityType type, boolean invert) {
+        addArgument("type", (invert ? "!" : "") + type.toString());
+        return this;
+    }
+
+    @Override
+    public Selector build() {
+        Preconditions.checkState(this.type != null);
+        boolean requiresLocation = this.arguments.containsKey("x") || this.arguments.containsKey("y") || this.arguments.containsKey("z") || this.arguments.containsKey("dx") || this.arguments.containsKey("dy") || this.arguments.containsKey("dz") || this.arguments.containsKey("rm") || this.arguments.containsKey("r") || (this.arguments.containsKey("c") && !this.type.equals(SelectorTypes.RANDOM_PLAYER));
```
kenzierocks: I'll double check this at some point.

src/main/java/org/spongepowered/api/text/selector/SpongeSelectorFactory.java
kenzierocks: Never mind, decimals aren't allowed/this matches them. It works in vanilla.

src/main/java/org/spongepowered/api/text/selector/SpongeSelectorFactory.java
kenzierocks: https://gist.github.com/kenzierocks/80d714dc1910719038c5

`getArgumentMap` is the replacement for your arg-parsing stuff, adapt it to your needs. the `main` function is just a test.

src/main/java/org/spongepowered/mod/text/selector/SpongeSelector.java
```diff
@@ -0,0 +1,92 @@
...
+@NonnullByDefault
+public class SpongeSelector implements Selector {
+
+    private final SelectorType type;
+    private final Map<String, String> arguments;
+    private final boolean requiresLocation;
+
+    public SpongeSelector(SelectorType type, Map<String, String> arguments, boolean requiresLocation) {
+        this.type = type;
+        this.arguments = arguments;
```
Minecrell: This has to be an immutable map, you can either do `ImmutableMap.copyOf(arguments)` here, or on in the builder's `build()` method before giving it to this constructor.

src/main/java/org/spongepowered/mod/text/selector/SpongeSelectorType.java
```diff
@@ -0,0 +1,44 @@
...
+ */
+package org.spongepowered.mod.text.selector;
+
+import org.spongepowered.api.text.selector.SelectorType;
+import org.spongepowered.api.util.annotation.NonnullByDefault;
+
+@NonnullByDefault
+public class SpongeSelectorType implements SelectorType {
+
+    private final String ID;
```
kenzierocks: Lowercase `id` instead of `ID`. All caps for `static final` only.

src/main/java/org/spongepowered/mod/text/selector/SpongeSelectorBuilder.java
```diff
@@ -0,0 +1,129 @@
...
+    @Override
+    public SelectorBuilder type(EntityType type, boolean invert) {
+        addArgument("type", (invert ? "!" : "") + type.toString());
+        return this;
+    }
+
+    @Override
+    public Selector build() {
+        Preconditions.checkState(this.type != null);
+        boolean requiresLocation = this.arguments.containsKey("x") || this.arguments.containsKey("y") || this.arguments.containsKey("z") || this.arguments.containsKey("dx") || this.arguments.containsKey("dy") || this.arguments.containsKey("dz") || this.arguments.containsKey("rm") || this.arguments.containsKey("r") || this.type.equals(SelectorTypes.NEAREST_PLAYER) || (this.arguments.containsKey("c") && !this.type.equals(SelectorTypes.RANDOM_PLAYER));
```
kenzierocks: Should be:
```boolean requiresLocation = this.arguments.containsKey("x") || this.arguments.containsKey("y") || this.arguments.containsKey("z") || this.arguments.containsKey("dx") || this.arguments.containsKey("dy") || this.arguments.containsKey("dz") || this.arguments.containsKey("rm") || this.arguments.containsKey("r") || this.type.equals(SelectorTypes.NEAREST_PLAYER);```

`c` doesn't matter for location. I'm thinking of a better way to do this checking that's more DRY.

kenzierocks: You can declare an `ImmutableList` holding the arguments that take location:
```private static final ImmutableList<String> LOCATION_DEPENDENT_ARGS = ImmutableList.of("x", "y", "z", "dx", "dy", "dz", "rm", "r");```
And use [`Collections.disjoint(Collection a, Collection b)`](http://docs.oracle.com/javase/6/docs/api/java/util/Collections.html#disjoint%28java.util.Collection,%20java.util.Collection%29) to determine the collision:
```boolean requiresLocation = !Collections.disjoin(this.arguments, LOCATION_DEPENDENT_ARGS) || this.type.equals(SelectorTypes.NEAREST_PLAYER);```
Which is much easier to read + maintain.

src/main/java/org/spongepowered/mod/text/selector/FakeCommandSender.java
kenzierocks: I see we do need it for resolving.

kenzierocks:
Lots of these classes contain no `toString()`, should that be defined?

src/main/java/org/spongepowered/mod/text/selector/SpongeSelectorBuilder.java
AIOOB: The loction is required if c is specified as `@a[c=1]` takes the closest playe, although it is not required if `@r` is used.

kenzierocks: Okay. Can you still switch to the other checking method I suggested, with `c` being a special case?

PR #107, opened 2015-01-01
Implement several more world methods

simon816:
I have tested all of these methods except `setSeed` and `setBiome`

There are some comments which it'd be good to have some input on

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
+++ b/src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
@@ -24,33 +24,35 @@
  */
 package org.spongepowered.mod.mixin.world;
 
-import java.io.File;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-import java.util.Random;
-import java.util.UUID;
+import static com.google.common.base.Preconditions.checkArgument;
+import static com.google.common.base.Preconditions.checkNotNull;
 
 import com.flowpowered.math.vector.Vector3d;
 import com.flowpowered.math.vector.Vector3i;
 import com.google.common.base.Optional;
-
-import static com.google.common.base.Preconditions.checkNotNull;
-import static com.google.common.base.Preconditions.checkArgument;
-
+import com.google.common.base.Predicate;
 import net.minecraft.network.Packet;
 import net.minecraft.profiler.Profiler;
 import net.minecraft.server.MinecraftServer;
 import net.minecraft.server.management.ServerConfigurationManager;
+import net.minecraft.util.BlockPos;
+import net.minecraft.world.ChunkCoordIntPair;
+import net.minecraft.world.GameRules;
 import net.minecraft.world.WorldProvider;
 import net.minecraft.world.WorldServer;
+import net.minecraft.world.biome.BiomeGenBase;
+import net.minecraft.world.chunk.IChunkProvider;
+import net.minecraft.world.gen.ChunkProviderServer;
 import net.minecraft.world.storage.ISaveHandler;
 import net.minecraft.world.storage.WorldInfo;
-
+import net.minecraftforge.common.ForgeChunkManager;
 import org.spongepowered.api.block.BlockLoc;
 import org.spongepowered.api.effect.particle.ParticleEffect;
+import org.spongepowered.api.effect.sound.SoundType;
 import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.EntitySnapshot;
 import org.spongepowered.api.entity.EntityType;
+import org.spongepowered.api.service.persistence.data.DataContainer;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.api.world.Chunk;
 import org.spongepowered.api.world.Dimension;
@@ -73,6 +75,15 @@ import org.spongepowered.mod.util.SpongeHooks;
 import org.spongepowered.mod.util.VecHelper;
 import org.spongepowered.mod.wrapper.BlockWrapper;
 
+import java.io.File;
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.Random;
+import java.util.UUID;
+
 @NonnullByDefault
 @Mixin(net.minecraft.world.World.class)
 public abstract class MixinWorld implements World, IMixinWorld {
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
```diff
@@ -93,21 +125,106 @@ public BlockLoc getBlock(int x, int y, int z) {
...
     @Override
     public Optional<Entity> createEntity(EntityType type, Vector3d position) {
+        try {
+            net.minecraft.entity.Entity entity =
+                    (net.minecraft.entity.Entity) type.getEntityClass().getConstructor(net.minecraft.world.World.class).newInstance(this);
+            entity.setPosition(position.getX(), position.getY(), position.getZ());
+            if (spawnEntityInWorld(entity)) {
+                return Optional.of((Entity) entity);
+            }
+        } catch (ReflectiveOperationException ignored) {
```
Minecrell: This doesn't compile, see https://travis-ci.org/SpongePowered/Sponge/builds/45614519#L514-L517

simon816: Oh, `ReflectiveOperationException` is Java 1.7, I'll change it

src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
```diff
@@ -93,21 +125,106 @@ public BlockLoc getBlock(int x, int y, int z) {
...
+    @Override
+    public void playSound(Sound sound, Vector3d position, double volume, double pitch, double minVolume) {
+        playSoundEffect(position.getX(), position.getY(), position.getZ(), sound.getName(), (float) Math.max(minVolume, volume), (float) pitch);
+    }
+
+    @Override
+    public Optional<Entity> getEntityFromUUID(UUID uuid) {
+        for (net.minecraft.entity.Entity entity : this.loadedEntityList) {
+            if (entity.getUniqueID().equals(uuid)) {
+                return Optional.of((Entity) entity);
```
Deamon5550: There is a better way to do this, WorldServer holds a Map of 'entitiesByUuid' which you can use to directly lookup the entity.

simon816: @Deamon5550 Done

simon816:
I have updated this PR to add a few more methods.
The subproject was updated in order to get the changes to the biome API, and so I have also fixed the WorldBorder mixin to change radius => border.

There are a few comments on some lines, I invite you to add github comments so I can make the most appropriate changes

src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
```diff
@@ -105,21 +138,54 @@ public BlockLoc getBlock(int x, int y, int z) {
...
+                    (net.minecraft.entity.Entity) type.getEntityClass().getConstructor(net.minecraft.world.World.class).newInstance(this);
+            entity.setPosition(position.getX(), position.getY(), position.getZ());
+            if (spawnEntityInWorld(entity)) {
+                return Optional.of((Entity) entity);
+            }
+        } catch (Exception ignored) {
+            // Entity constructor does not have single World parameter, too bad
+        }
         return Optional.absent();
     }
```
AIOOB: It's probably safer to use `EntityList.createEntityByName(String, World)`

simon816: I suppose it might be, but since we already know the class (`type.getEntityClass()`) it's probably quicker to use that than to use a map lookup

src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
```diff
@@ -260,4 +328,94 @@ public boolean doesKeepSpawnLoaded() {
...
+            return ((ChunkProviderServer) provider).loadedChunks;
+        }
+        return Collections.EMPTY_LIST;
+    }
+
+    @Override
+    public boolean deleteChunk(Chunk chunk) {
+        IChunkProvider provider = this.getChunkProvider();
+        Vector3i chunkPos = chunk.getPosition();
+        if (provider instanceof ChunkProviderServer) {
```
bloodmc: Every worldserver creates a ChunkProviderServer instance, there is no need to check anything else. There also needs to be a check for forced chunks as many forge mods contain blocks that keep chunks loaded. If a forced chunk is found, this method needs to return false.

src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
```diff
@@ -260,4 +328,94 @@ public boolean doesKeepSpawnLoaded() {
...
+
+    @Override
+    public void setSeed(long seed) {
+        ObfuscationReflectionHelper.setPrivateValue(WorldInfo.class, this.worldInfo, seed, "randomSeed");
+        this.rand.setSeed(seed);
+    }
+
+    @SuppressWarnings("unchecked")
+    @Override
+    public Iterable<Chunk> getLoadedChunks() {
```
bloodmc: Same issue as below, this method only needs to return the loaded chunks of the ChunkProviderServer.

src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
```diff
@@ -260,4 +328,94 @@ public boolean doesKeepSpawnLoaded() {
...
+    }
+
+    @Override
+    public long getWorldSeed() {
+        return this.getSeed();
+    }
+
+    @Override
+    public void setSeed(long seed) {
+        ObfuscationReflectionHelper.setPrivateValue(WorldInfo.class, this.worldInfo, seed, "randomSeed");
```
progwml6: @bloodmc do we want to use AT's for things like this to prevent errors during MC version updates?

bloodmc: Yes, this should be added to our sponge_at.cfg

src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
```diff
@@ -126,21 +159,56 @@ public BlockLoc getBlock(int x, int y, int z) {
...
     @Override
     public BiomeType getBiome(Vector3i position) {
-        throw new UnsupportedOperationException();
+        return (BiomeType) this.getBiomeGenForCoords(VecHelper.toBlockPos(position));
     }
 
     @Override
     public void setBiome(Vector3i position, BiomeType biome) {
-        // TODO
+        net.minecraft.world.chunk.Chunk chunk = this.getChunkFromBlockCoords(VecHelper.toBlockPos(position));
```
bloodmc: This needs a check for null.

src/main/java/org/spongepowered/mod/mixin/world/MixinWorld.java
```diff
@@ -288,4 +361,88 @@ public void setKeepSpawnLoaded(boolean keepLoaded) {
...
+
+    @Override
+    public void setGameRule(String gameRule, String value) {
+        this.worldInfo.getGameRulesInstance().setOrCreateGameRule(gameRule, value);
+    }
+
+    @Override
+    public Map<String, String> getGameRules() {
+        GameRules gameRules = this.worldInfo.getGameRulesInstance();
+        // TODO Could expose GameRules#theGameRules via reflection to remove the
```
Zidane: Just make this public in the AT...

simon816: OK, so there's no problem with exposing the underlying Map I take it.

simon816: Actually, there's not really much advantage of doing that so I'll remove the comment.

PR #108, opened 2015-01-01
Implement armor stand entity and ArmorEquipable

simon816:
Only a few methods tested but the methods are rather repetitive in their function so should work for all.

There's some issues with methods calling themselves due to same signature as shadowed method, so comment those methods out in development environment until @Mumfrey has a way to fix this.


Mumfrey:
> so comment those methods out in development environment until @Mumfrey has a way to fix this.

There already is a way to fix this, it's [explained at length in the mixin examples](https://github.com/SpongePowered/Sponge/blob/master/src/example/java/org/spongepowered/mod/mixin/MixinEntityPlayerExample.java).

simon816:
Ahh yes, forgot about `@Overrite`, changed it

simon816:
I have updated this PR to make use of a mixin's ability to target multiple classes. MixinArmorEquipable attaches onto all entity classes that have an ArmorEquipable interface. This is a useful technique to avoid copying the same implementation from mixin to mixin.

src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -555,7 +555,7 @@ private void setEntityTypes() {
         this.entityTypeMappings.put("PRIMED_TNT", new SpongeEntityType(20, "PrimedTnt", (Class<? extends Entity>)EntityList.stringToClassMapping.get("PrimedTnt")));
         this.entityTypeMappings.put("FALLING_BLOCK", new SpongeEntityType(21, "FallingSand", (Class<? extends Entity>)EntityList.stringToClassMapping.get("FallingSand")));
         this.entityTypeMappings.put("FIREWORK", new SpongeEntityType(22, "FireworksRocketEntity", (Class<? extends Entity>)EntityList.stringToClassMapping.get("FireworksRocketEntity")));
-        this.entityTypeMappings.put("ARMORSTAND", new SpongeEntityType(30, "ArmorStand", (Class<? extends Entity>)EntityList.stringToClassMapping.get("ArmorStand")));;
+        this.entityTypeMappings.put("ARMOR_STAND", new SpongeEntityType(30, "ArmorStand", (Class<? extends Entity>)EntityList.stringToClassMapping.get("ArmorStand")));;
```
kenzierocks: Extra semicolon.

simon816: Was already like that. I'll leave it because there's not much point making one minor fix to the file that has lots of minor issues. I will fix this in https://github.com/SpongePowered/Sponge/pull/112 if I push there again

bloodmc:
Merged thanks!

PR #109, opened 2015-01-02
Implement ItemStack, TradeOffer and PotionEffect builders

simon816:
The required mixins for ItemStack, TradeOffer and PotionEffect are implemented, but there was no builders for them.
This PR implements the 3 builders.

### DIFF
--- a/src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
+++ b/src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
@@ -25,13 +25,12 @@
 
 package org.spongepowered.mod.registry;
 
-import java.awt.Color;
-import java.lang.reflect.Field;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
+import com.google.common.base.Optional;
+import com.google.common.collect.ImmutableBiMap;
+import com.google.common.collect.ImmutableBiMap.Builder;
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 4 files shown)


spaceemotion:
Just curious - since the travis build failed - does the project have any kind of words to say about unit testing? Whether to add it in the API, the Forge implementation or elsewhere? That it needs to be present in PRs?

Would be interesting to know :)

simon816:
Oops I forgot to update the API submodule, hence the build failed, will fix.

A unit test for builders would be a good idea

src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -114,18 +113,22 @@
...
 import org.spongepowered.mod.text.format.SpongeTextColor;
 import org.spongepowered.mod.weather.SpongeWeather;
 import org.spongepowered.mod.world.SpongeDimensionType;
 
-import com.google.common.base.Optional;
-import com.google.common.collect.ImmutableBiMap;
-import com.google.common.collect.ImmutableBiMap.Builder;
-import com.google.common.collect.ImmutableList;
-import com.google.common.collect.ImmutableMap;
-import com.google.common.collect.Maps;
```
gabizou: This sounds a bit like OCD fixes. Is this necessary?

simon816: Not necessary but since this file is being modified I thought it might as well be done at the same time

bloodmc:
Merged thanks.

PR #111, opened 2015-01-03
Implement text actions

AIOOB:
The TextAction id needs to be revised, I couldn't work out what it should be.


src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -872,6 +874,14 @@ private void setWeathers() {
         }
     }
 
+    private void setTextActionFactory() {
+        try {
+            TextActions.class.getDeclaredField("factory").set(null, new SpongeTextActionFactory());
```
Mumfrey: Wouldn't this require ```setAccessible(true)``` in order to work? Surely it'll just fail silently because you're catching the exception?

AIOOB: The line added to the AT makes it accessible.

Mumfrey: Ah alrighty then.

bloodmc:
Merged thanks.

PR #112, opened 2015-01-03
Several improvements to SpongeGameRegistry

simon816:
* All mapping fields are final
 * directionMap follows other static builder conventions
 * RegistryHelper - makes mapping API classes easier
  * `Map<String, ?>` mapping like before. Optional ignoredFields parameter
  * `Function<String, ?>` mapping, calls a function for every field in the target class
  * `setFactory` convenience method
 * `newEntityTypeFromName` - simplifies entity mapping
  * Looks up entity ID, no need to know it before hand
  * Reduces clutter, can spot mistakes easier. e.g Who spotted `WITCH` => `FallingSand`
 * In `setBiomeTypes`, reduced the amount of calls to `BiomeGenBase.getBiomeGenArray()` by making local variable `biomeArray`

### Registry issues fixed with this PR
 * ~~Entity type mapping `ARMOR_STAND` was named `ARMORSTAND`~~ Has been fixed
 * Entity type mapping `WITCH` was actually mapped to `FallingSand`
 * The `entityIdToTypeMappings` was actually setting `modId` (i.e `'minecraft'`) as the key rather than `getId()`

### DIFF
--- a/src/main/java/org/spongepowered/mod/registry/RegistryHelper.java
+++ b/src/main/java/org/spongepowered/mod/registry/RegistryHelper.java
@@ -1,0 +1,79 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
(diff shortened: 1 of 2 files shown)


simon816:
Pushed an update to add selector factories using `setFactory` and remove unnecessary changes to biome mapping and imports

bloodmc:
Merged thanks!

PR #113, opened 2015-01-04
Implement Human interface and remainder of methods in Player interface

mproncace:
This PR adds the MixinEntityPlayer class, which implements methods defined in the Human interface. Additionally, it implements the clearTitle and resetTitle methods defined in the Player interface.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/entity/player/MixinEntityPlayer.java
+++ b/src/main/java/org/spongepowered/mod/mixin/entity/player/MixinEntityPlayer.java
@@ -1,0 +1,75 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/mod/mixin/entity/player/MixinEntityPlayer.java
```diff
@@ -0,0 +1,75 @@
...
+import org.spongepowered.api.entity.living.Human;
+import org.spongepowered.api.util.annotation.NonnullByDefault;
+import org.spongepowered.asm.mixin.Implements;
+import org.spongepowered.asm.mixin.Interface;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+@NonnullByDefault
+@Mixin(EntityPlayer.class)
+@Implements(@Interface(iface = Human.class, prefix = "human$"))
```
Mumfrey: Why is this a soft implements? I can't see any conflicts with methods in the target class?

simon816: It's conflicts in `EntityLivingBase`, the same reason my armor stand mixin is soft implements

Mumfrey: Oh it's a superinterface method that conflicts? That makes sense then.

bloodmc:
Merged thanks!

PR #114, opened 2015-01-04
Implement GameMode

Aaron1011:


src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -167,6 +171,14 @@
                                                                            .put("LEFT", new SpongeRotation(270))
                                                                            .put("TOP_LEFT", new SpongeRotation(315))
                                                                            .build();
+    private static final ImmutableMap<String, GameMode> gameModeMappings =  new ImmutableMap.Builder<String, GameMode>()
```
bloodmc: You are missing "NONE". FakePlayers have this gamemode initially in Forge.

Aaron1011: But, that's never going to get used here.

`setGameModes` iterates over the fields in `GameModes`, and uses the name of the field as the key. Unless `NOT_SET` is added to `GameModes`, adding it here does nothing.

bloodmc: Make a PR that fixes it in API as well.

Aaron1011:
This is now unblocked, as my corresponding API PR has been merged.

src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -207,6 +218,34 @@
...
+        Builder<Direction, EnumFacing> directionMapBuilder = ImmutableBiMap.builder();
+        directionMapBuilder.put(Direction.NORTH, EnumFacing.NORTH);
+        directionMapBuilder.put(Direction.EAST, EnumFacing.EAST);
+        directionMapBuilder.put(Direction.SOUTH, EnumFacing.SOUTH);
+        directionMapBuilder.put(Direction.WEST, EnumFacing.WEST);
+        directionMapBuilder.put(Direction.UP, EnumFacing.UP);
+        directionMapBuilder.put(Direction.DOWN, EnumFacing.DOWN);
+        directionMap = directionMapBuilder.build();
+    }
+>>>>>>> Implement GameMode
```
simon816: Merge conflict dump left here.

src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -950,6 +959,16 @@ private void setDimensionTypes() {
         }
     }
 
+    private void setGameModes() {
+        for (Field f: GameModes.class.getDeclaredFields()) {
+            try {
+                f.set(null, gameModeMappings.get(f.getName()));
```
simon816: You can now use [`RegistryHelper`](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/registry/RegistryHelper.java) here

src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
simon816: @Aaron1011 `NOT_SET` can be added now since it's in the API

src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -950,6 +959,16 @@ private void setDimensionTypes() {
...
+            } catch (Exception e) {
+                e.printStackTrace();
```
progwml6: we should be logging exceptions NOT using printStackTrace

simon816: @progwml6 That should be done in [RegistryHelper](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/registry/RegistryHelper.java) then :p

PR #116, opened 2015-01-05
Implement Bat entity

mproncace:
This PR implements the Bat entity in MixinEntityBat.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityBat.java
+++ b/src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityBat.java
@@ -1,0 +1,60 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityBat.java
```diff
@@ -0,0 +1,60 @@
...
+
+    @Shadow
+    public abstract void setIsBatHanging(boolean p_82236_1_);
+
+    public MixinEntityBat(World worldIn) {
+        super(worldIn);
+    }
+
+    public boolean bat$isAwake() {
+        return getIsBatHanging();
```
simon816: Should be `!getIsBatHanging()` since the method is the opposite

src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityBat.java
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public void bat$setAwake(boolean awake) {
+        setIsBatHanging(awake);
```
simon816: `setIsBatHanging(!awake)` for same reason as above

src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityBat.java
mproncace: Corrected.

bloodmc:
Merged thanks!

PR #123, opened 2015-01-06
Implement boat entity

mproncace:
This commit implements the boat entity and injects necessary vanilla modifications to support API functions.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityBoat.java
+++ b/src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityBoat.java
@@ -1,0 +1,122 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.mod.mixin.entity.living;
+
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.entity.item.EntityBoat;
+import net.minecraft.util.AxisAlignedBB;
+import net.minecraft.world.World;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityBoat.java
```diff
@@ -0,0 +1,122 @@
...
+    @Inject(method = "onUpdate()V", at = @At(value = "INVOKE", target = "net.minecraft.entity.Entity.moveEntity(DDD)V"))
+    public void implementLandBoats(CallbackInfo ci){
+        if (this.onGround && this.moveOnLand){
+            this.motionX /= 0.5D;
+            this.motionY /= 0.5D;
+            this.motionZ /= 0.5D;
+        }
+    }
+
+    // this code is injected three times, but the first conditional implicitly fails at the second two locations
```
Deamon5550: Rather than having this inject three times it would be better to use [At#ordinal()](https://github.com/SpongePowered/Mixin/blob/master/src/main/java/org/spongepowered/asm/mixin/injection/At.java#L99).

mproncace: Ah, wasn't aware of that. I'll push a new commit momentarily.

src/main/java/org/spongepowered/mod/mixin/entity/vehicle/MixinEntityBoat.java
```diff
@@ -0,0 +1,121 @@
...
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+@NonnullByDefault
+@Mixin(EntityBoat.class)
+public abstract class MixinEntityBoat extends Entity implements Boat {
+
+    private double maxSpeed = 0.4D;
```
simon816: This won't work IIRC, since field initializers are dropped by the mixin transformer. Only static fields can have an initial value.

mproncace: Are the fields themselves dropped, or can I just assign the initial values in the vanilla constructor?

mproncace: After looking at Mixin's source, it looks like the fields will be mixed in, just not initialized. I'll move initialization to the constructor in that case.

simon816: Unfortunately constructors don't get mixed in, for the same reason instance field initializers don't either.
I cannot think how to overcome these limitations, it may be possible to `@Inject` on a constructor (`<init>`) though I doubt it.
How could this be done @Mumfrey?

Mumfrey: You can `@Inject` into a constructor as long as you use "RETURN" as the `@At` type.

ryantheleach: https://github.com/SpongePowered/Sponge/blob/master/src/example/java/org/spongepowered/mod/mixin/MixinWorldProviderExample.java#L63

You can inject on a constructor, but it can only run right at the end before it returns, so you could set the field but it wouldn't be an initialization of the field but I suspect it will be just a normal set and might cause problems if the field is final. @Mumfrey will need to confirm.

kenzierocks: Bytecode allows setting of the classes final anywhere in that class, so it should be fine.

Mumfrey: Yes but java doesn't, so how would you compile such a class?

kenzierocks: More ASM magic? This field isn't final right now anyways.

Mumfrey: I know it's not final, but you were talking about final fields. You could set a final shadow field but I'm saying if you put a field in a mixin and made it final then there's no way to compile code which sets the value of it (other than using reflection), ASM won't fix the fact that the class wouldn't compile in the first place.

mproncace: This discussion is irrelevant, as the PR doesn't deal with final fields at all.

bloodmc:
Merged thanks!

PR #125, opened 2015-01-06
Implement User interface

mproncace:
This PR implements SpongeAPI's `User` interface within the `SpongeUser` wrapper class.

### DIFF
--- a/src/main/java/org/spongepowered/mod/entity/SpongeUser.java
+++ b/src/main/java/org/spongepowered/mod/entity/SpongeUser.java
@@ -1,0 +1,117 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.mod.entity;
+
+import com.google.common.base.Optional;
+import com.mojang.authlib.GameProfile;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.server.MinecraftServer;
+import net.minecraft.world.World;
+import net.minecraft.world.storage.SaveHandler;
+import net.minecraftforge.fml.common.FMLCommonHandler;
+import org.spongepowered.api.Server;
+import org.spongepowered.api.entity.player.Player;
+import org.spongepowered.api.entity.player.User;
+import org.spongepowered.mod.SpongeMod;
+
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/mod/entity/SpongeUser.java
```diff
@@ -0,0 +1,117 @@
...
+
+    public boolean hasJoinedBefore() {
+        return this.getData().isPresent();
+    }
+
+    @Override
+    public Date getFirstPlayed() {
+        @Nullable NBTTagCompound data = getData().isPresent() ? getData().get() : null;
+        long time = 0L;
+        if (data != null && data.hasKey("firstPlayed")) {
```
Sannholm: Why not just do:
```
Optional<NBTTagCompound> data = getData();
if(data.isPresent() && data.get().hasKey("firstPlayed")) {...}
```

src/main/java/org/spongepowered/mod/entity/SpongeUser.java
```diff
@@ -0,0 +1,117 @@
...
+        Calendar cal = Calendar.getInstance();
+        cal.setTimeInMillis(time);
+        return cal.getTime();
+    }
+
+    @Override
+    public Date getLastPlayed() {
+        @Nullable NBTTagCompound data = getData().isPresent() ? getData().get() : null;
+        long time = 0L;
+        if (data != null && data.hasKey("lastPlayed")) {
```
Sannholm: Same as above here

src/main/java/org/spongepowered/mod/entity/SpongeUser.java
```diff
@@ -0,0 +1,117 @@
...
+    public SpongeUser(World worldIn, GameProfile profile) {
+        super(worldIn, profile);
+    }
+
+    @Override
+    public String getName() {
+        return getGameProfile().getName();
+    }
+
+    public boolean hasJoinedBefore() {
```
Mumfrey: Missing ```@Override``` here.

src/main/java/org/spongepowered/mod/entity/SpongeUser.java
```diff
@@ -0,0 +1,117 @@
...
+        return server.getConfigurationManager().getBannedPlayers().isBanned(getGameProfile());
+    }
+
+    @Override
+    public boolean isWhitelisted() {
+        MinecraftServer server = FMLCommonHandler.instance().getMinecraftServerInstance();
+        return server.getConfigurationManager().getWhitelistedPlayers().func_152705_a(getGameProfile());
+    }
+
+    public boolean isOnline() {
```
Mumfrey: Also missing ```@Override``` here

src/main/java/org/spongepowered/mod/mixin/entity/player/MixinEntityPlayerMP.java
```diff
@@ -55,7 +55,7 @@
 @NonnullByDefault
 @Mixin(EntityPlayerMP.class)
 @Implements(@Interface(iface = Player.class, prefix = "playermp$"))
-public abstract class MixinEntityPlayerMP extends EntityPlayer implements CommandSource {
+public abstract class MixinEntityPlayerMP extends SpongeUser implements CommandSource {
```
Mumfrey: No, this makes no sense.

src/main/java/org/spongepowered/mod/entity/SpongeUser.java
mproncace: You're right, that would probably make a bit more sense.

src/main/java/org/spongepowered/mod/entity/SpongeUser.java
mproncace: Fixed.

src/main/java/org/spongepowered/mod/mixin/entity/player/MixinEntityPlayerMP.java
mproncace: It still indirectly extends EntityPlayer via the new wrapper.

Mumfrey: This is a **mixin**, it's not a class! You cannot just arbitrarily change the parent.

mproncace: Ah, okay. It won't cause any new issues if I switch it back, then?

Mumfrey: It will cause issues if you **don't** switch it back, the mixin will not work if you arbitrarily change its superclass.

mproncace: Understood, but will methods from `User` still be implemented for `EntityPlayerMPs`?

Mumfrey: Implemented? No, because you've implemented the ```User``` interface in a separate class which doesn't appear anywhere in the hierarchy of the ```Player``` implementation (mixed in to ```EntityPlayerMP```). I'm not 100% sure what you're actually trying to do here to be honest.

mproncace: I was going off a suggestion from [this issue](https://github.com/SpongePowered/Sponge/issues/122). I'm going to close this PR and find a better solution, though.

PR #126, opened 2015-01-06
Implement minecart entities

caseif:
This PR implements the Minecart entity along with all of its subinterfaces.


src/main/java/org/spongepowered/mod/mixin/entity/vehicle/minecart/MixinEntityMinecart.java
```diff
@@ -0,0 +1,138 @@
...
+            this.motionY *= 0.0D;
+            this.motionZ *= 0.96D;
+        }
+    }
+
+    public MixinEntityMinecart(World worldIn) {
+        super(worldIn);
+    }
+
+    public double minecart$getMaxSpeed() {
```
Mumfrey: This is going to overwrite the forge method, which probably isn't want you wanted.

mproncace: Hm, I had thought a soft-implementation would prevent that. Should I transform the Forge method, then?

Mumfrey: The soft implementation allows things which are valid in **bytecode** but aren't valid in **java code** to coexist by running them through the compiler with one name and then changing it (stripping the prefix) at run time. However since the method still gets injected with the same name, it will overwrite the target just like any other mixin method would.

You have two choices really:

1. Do nothing, just remove this mixin method. This is fine because the forge method will happily "implement" the interface.
2. If you need different behaviour, for example if you want the value set for ```maxSpeed``` to act like a limit to the underlying value, then use an injector. Injectors using RETURN can actually read and change the original return value so for example:
 
 ```java
@Inject(target = "getMaxSpeed()D", at = @At("RETURN"), cancellable = true)
public double onGetMaxSpeed(CallbackInfoReturnable ci) {
        double originalValue = ci.getReturnValue();
        ci.setReturnValue(Math.min(originalValue, this.maxSpeed));
}
```

Comment (mproncace) [inline src/main/java/org/spongepowered/mod/mixin/entity/vehicle/minecart/MixinEntityMinecart.java]: @Mumfrey The issue here is that Sponge's `getMaxSpeed()` is expected to return a constant value (whatever `maxSpeed` is currently set to), whereas Forge's takes other factors into account meaning it will vary depending on when you call it. To put it more simply, they're fundamentally two different methods and need to coexist.

File: src/main/java/org/spongepowered/mod/mixin/entity/vehicle/minecart/MixinEntityMinecartCommandBlock.java [hunk 2]
```diff
@@ -0,0 +1,71 @@
...
+import org.spongepowered.asm.mixin.Interface;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+@NonnullByDefault
+@Mixin(EntityMinecartCommandBlock.class)
+@Implements(@Interface(iface = MinecartCommandBlock.class, prefix = "commandminecartminecart$"))
```
simon816: `cmdCart$` would be a smaller prefix

src/main/java/org/spongepowered/mod/mixin/entity/vehicle/minecart/MixinEntityMinecart.java
```diff
@@ -0,0 +1,138 @@
...
+    @Shadow
+    public abstract double getDragAir();
+
+    @Shadow
+    public abstract boolean isOnGround();
+
+    private double maxSpeed = 0.4D;
+    private boolean slowWhenEmpty = true;
+    private Vector3d airborneMod = new Vector3d(0.5D, 0.5D, 0.5D);
+    private Vector3d derailedMod = new Vector3d(0.5D, 0.5D, 0.5D);
```
simon816: These fields won't be initialized, same issue as [this comment](https://github.com/SpongePowered/Sponge/pull/123/files#r22581632)

caseif:
The Travis build will pass when the SpongeAPI submodule is updated.

caseif:
The Travis build will pass when built against a newer SpongeAPI revision.

src/main/java/org/spongepowered/mod/mixin/entity/vehicle/minecart/MixinEntityMinecartFurnace.java
```diff
@@ -0,0 +1,53 @@
...
+public abstract class MixinEntityMinecartFurnace extends EntityMinecart implements MinecartFurnace {
+
+    @Shadow
+    private int fuel;
+
+    public MixinEntityMinecartFurnace(World worldIn) {
+        super(worldIn);
+    }
+
+    public int furnaceminecart$getFuel() {
```
simon816: Prefixes have been left here since changing to hard implements, needs removing. (also remember to put `@Override`)

modwizcode: @simon816 can you clarify what you mean?

Mumfrey: @modwizcode it's a prefixed method presumably from a soft implements but the implementation is hard, so the prefix should not be there.

bloodmc:
@caseif Can you rebase this PR and make sure it builds cleanly with Travis.

src/main/java/org/spongepowered/mod/mixin/entity/vehicle/minecart/MixinEntityMinecartCommandBlock.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    @Override
+    public String getCommandName() {
+        return field_145824_a.getCustomName();
+    }
+
+    @Override
+    public void setCommandName(@Nullable String name) {
+        if (name == null)
```
simon816: Should have braces `{}`, even if it's only a one-line if block

src/main/java/org/spongepowered/mod/mixin/entity/vehicle/minecart/MixinEntityMinecart.java
```diff
@@ -0,0 +1,154 @@
...
+    @Override
+    public Vector3d getDerailedVelocityMod() {
+        return this.derailedMod;
+    }
+
+    @Override
+    public void setDerailedVelocityMod(Vector3d derailedVelocityMod) {
+        this.derailedMod = derailedVelocityMod;
+    }
+}
```
simon816: Missing newline

PR #128, opened 2015-01-07
Feature/scheduler

sibomots:
This implementation tries to solve the problem of providing the capability to create Tasks on behalf of the Plugin and execute the tasks per the parameters of the Task.

A Task is simply created by the user-developer in the plugin as follows:

1.  The user-developer makes a Runnable thread.  It's preferred that the Runnable Thread will exit or at the least manage itself carefully (We'll cover the issues about that later).

For example, let's suppose a list of tasks in a Plugin is maintained:

```java
    Thread task = new Thread(new Runnable() {
       private long counter = 0;
       public void run() {
           System.out.println(" Task  inside Plugin runs [" + counter + "]");
               counter++;
           }
       });
```

Then suppose the tasks were setup through the Event handler in the plugin for `ServerStartedEvent`

```java
    @Subscribe
    public void handler(ServerStartedEvent evt) {
        game = evt.getGame();

        Optional<PluginContainer> result = evt.getGame().getPluginManager().getPlugin("MyPluginID");
        PluginContainer pluginContainer = result.get();

        Scheduler scheduler = game.getScheduler();

        // Make the Task task run with an initial offset delay of 500 ticks, and make the period of
        // of the Task to repeat every 2000 ticks

        scheduler.runRepeatingTaskAfter(pluginContainer, task, 500, 2000);
       
    }
```

There are other ways to create Tasks.  Some are one-time-shot, some are repeated.  Both types can have an offset to wait before occurring.  And Repeating Tasks have period > 0.

The Scheduler does a simple job of making sure that when Tasks repeat, the next time the task should occur, the Task will be skipped if the previous occurrence of the Task had not finished.  When this happens a warning log message will be sent to the Logger.

Other Logging messages for other warning/error conditions also can be seen in the log when conditions arise.

The Scheduler uses a simple wait/notify scheme to yield the CPU so that the Scheduler treads as lightly as possible on the system.

There are a couple missing methods so far: To fetch the list of Tasks and fetch the list of tasks that are owned by a particular plugin.  I'm going to add those later after I decide how to treat the persistence of the Task data.

I've tested the Scheduler lightly with a simple Plugin and it seems to work as designed.  I could use comments and critique in the peer review before expecting the code to be merged in the main branch.

I've tried to put enough detail in comments and JavaDoc to elaborate the design notes and usage.

Questions/comments/concerns -- just email me.  

Thanks.


sibomots:
On the bus ride home from work some other thoughts occurred to me about the implementation of the Scheduler.

First, let me point out that the API was designed by another developer and although I don't think the initial design of the API is faulty, there are some minor changes that would make the overall behavior of the Scheduler a bit safer and more reliable.

Let me summarize some of the comments that I've read thus far.

One comment was along the lines that the API signature of the Scheduler shouldn't involve the use of a Runnable (Thread) argument.  The notion of the argument against using Runnable was that the Scheduler itself internally should be in charge of managing _how_ a Task is implemented and not expose the raw Thread in the Plugin code.   I agree and initially before I had decided to stick to the API I had had devised that the API signature of the Scheduler would involve the passing in of Method not Runnable. The Scheduler would synthesize a Thread, call the "Method" (loosely think call-back metaphor) and run the Method.  That tactic doesn't really solve the whole problem -- a beligerent user-developer can still write a Method that runs too long for a RepeatingTask, or causes other errors to occur that deal with the problem of concurrency.    

The other comment raised was to extend Runnable and create a specific Sponge Task thread internally to the scheduler which would have some behavior to make it safer.  For example, if a Task was requested, the Scheduler would concoct a new SpongeTaskThread (for lack of a better name) and within it has a built in timer or some mechanism to avoid running past a certain point (either the duration of period between repeating Tasks, or some other configurable/tuneable limit).

I'm not against the idea of that solution except that it makes the execution of the Task even more arbitrary than before and it may confuse or frustrate user-developers who want full expectation of a free-running Task.

There were a couple methods that I had decided were unnecessary.  One was the `getName()` method for a Task.  Where would that be useful?  I wonder.

The other two methods that I think we do need, but are not yet implemented are the two methods to  get the list of Tasks (either keyed by Plugin owner) or the global list of Tasks.  I'm on the fence where this is useful, but I will have to discuss with folks on IRC/here to find out more background.

Some minor nits to pick either way (pro or con):  The methods to get the total duration, total period, or remaining time left before the pending Task will run.  Those could be useful.  Perhaps I can add them to the API based on comments here.

The goal from the start was to create a Scheduler that was synchronous to a timed event in the game and the ServerTick was chosen for that purpose.  It gives user-developers a way to attempt to have a Task run as near as possible to the rising edge of a TickEvent.  I haven't yet added any code to measure latency in real-time (which would be interesting from a performance API perspective).

I'm unsure why the code that was checked in fails Travis; it seems that the API checkin isn't being picked up during the Travis BVT.  I'll ask the fellas about that later on.

Thanks.

ryantheleach:
Would taking a Method for the scheduler API work with java 8 method references? http://docs.oracle.com/javase/tutorial/java/javaOO/methodreferences.html

sibomots:
@ryantheleach I don't think that taking Method arguments would fail.  But the Java 8 feature of lambda expressions would break the build since we're choosing to baseline all code up to the Java 1.6 (IIRC) capability.  If sometime down the line Minecraft, Forge and thus Sponge decide to elevate the JDK to Java 8 then that would be a nicer feature to mingle with the Sponge API.  In the short term as we prepare release of Sponge I just don't see it happening.

It is an interesting option though.

sibomots:
Removing this botched branch/PR due to my mistakes with git.  Re-created under branch feature/syncscheduler (passed BVT)

PR #129, opened 2015-01-08
Implement Arrow

Aaron1011:


src/main/java/org/spongepowered/mod/mixin/entity/projectile/MixinEntityArrow.java
```diff
@@ -0,0 +1,104 @@
...
+    @Shadow
+    public Entity shootingEntity;
+
+    // Not all ProjectileSources are entities (e.g. BlockProjectileSource).
+    // This field is used to store a ProjectileSource that isn't an entity.
+    @Nullable
+    public ProjectileSource projectileSource;
+
+    @Override
+    public ProjectileSource getShooter() {
```
Aaron1011: This method, along with `setShooter`, is going to be duplicated in several other projectile impls. However, I'm not sure if there's a good way to pull it out of this class.

simon816: I had the same thoughts with ArmorEquipable. You can create a mixin that targets multiple classes so you could create a MixinProjectile targeting all implementers of Projectile. See [MixinArmorEquipable](https://github.com/SpongePowered/Sponge/pull/108/files#diff-0) for how this could be done.

Aaron1011: After looking the NSM projectile classes, it turns out that I'll be able to remove *some* duplication, but not all of it.

Some projectiles extend `EntityThrowable`, which contains logic for handling the shooter. However, for classes like `EntityArrow` which *don't*, I'll have to copy-paste it with simple changes, despite the logic being the same.

bloodmc:
Merged thanks!

PR #130, opened 2015-01-08
Feature/syncscheduler

sibomots:
The PR attempts to solve the problem of having a synchronous scheduler in the Sponge.

How the Scheduler works:

Upon the rising edge of the signal ServerTickEvent (Phase.START), a counter is incremented.  This is the basic time keeping mechanism of the scheduler.  When the SpongeMod has competed its own PreInit phase, the Scheduler has been started.

Each user-developer Plugin may create  Tasks.  A Task can be be a "one time shot" or a repeating Task.

The type of Task, and the parameters of how the Task runs are passed in the method call when the Task is created.  For example:

```java
public Optional<Task> runTaskAfter(Object plugin, Runnable task, long delay)
```
The method creates a Task that will run after *delay* ticks.

The parameters for creating a task can specify:

The delay  (offset in ticks from the moment the task is created until it is first run)
The interval (the period between occurrences of the Task running.)

Neither Delay nor Interval can be modified after the Task is created. 

Tasks can be canceled by using the Task interface ```cancel()```.

When a repeating Task has been canceled, it will not occur the next time it would have been run.  All canceled Tasks are removed from the list of Scheduled Tasks.  There is no undo for ```cancel()```.

A Task may be one-time shot or repeating based on the Interval of the Task (that and the method used to create the Task is the more obvious hint of which kind of Task is created).   In any event, if querying the Task the Interval is absent, then the Task has no Interval, thus is non-Repeating.  Only Tasks with Intervals are repeating.

A task may have no delay.  When querying a Task, if the delay is absent, then the Task was created without a delay.

Once a Task has begun, the task occurrence cannot be canceled.  That occurrence of the Task must complete.  One-time shot Tasks (non-repeating) have to be canceled before they run.   A repeating Task may run multiple times until it is canceled.  The cancellation of a repeating Task only stops any new occurrences from happening.

Synchronicity

The event that is listened by the Scheduler is the ServerTickEvent.  Only on the Phase.START phase of the event will the Scheduler's own internal counter be incremented and any pending Tasks be checked.

This has one side-effect:  Tasks will run in the same thread context as the thread which caused the ServerTickEvent.  This means that other TickEvent handlers that occur after will block until the Tasks (due for execution) are run.  This is how the Scheduler makes sure that the world data (potentially) modified by Plugins is coherent with the state of the game.

Unfortunately, the actual period of Tasks being checked is not monotonous.  The ServerTickEvent is subordinate to the activity of the Server and there are instances where the event will be late or early.  This means that the notion of "next Tick" or monotonous tick counting is nearly impossible because the Tick is not based on Wall-Clock timer but on a synthetic timer in the Server every 50ms.  However, the ServerTickEvent is not guarenteed to occur every 50ms.  

For users seeking a more regular and accurate Timed Task Scheduler, an Asynchronous Scheduler is better suited.  It will not be synchronous to the ServerTickEvent, but the timeliness of each AsynchronousTask is based on a real clock in a independent thread.

There are copious notes in the Implementation and JavaDoc elaborations there to read.  

Questions/Comments/Concerns -- feel free to drop me a line here or directly.

Thanks for your help.


Zidane:
@sibomots 

How complex do you think it would be to be able to determine if your task runs at the head of the tick or the end of it?

Also, its worth nothing that FML's tick event also fires for different object types. There are RENDER, WORLD, SERVER, PLAYER, CLIENT...obviously a few of those don't apply to Sponge.

sibomots:
```onTick``` handler in the Scheduler blocks when the ```ServerTickEvent``` fires. If the said event is ```Phase.START```, then all possible ```ScheduledTask```s run, so to answer 1st question.  prima facia - All run upon rising edge of ```ServerTickEvent```.

I could instrument the handler to make time measurements when the last Task is run and when the ```onTick``` event handler fires to see how close to the ```Phase.END``` the Task actually ran.

Zidane:
I brought it up as a topic of discussion as the only reason I can think of one wanting START tasks and END tasks is for a dev who feels they need to run a Task after another task runs in the same commit.

Zidane:
My comment regarding types is simply the fact that do we want tasks running on any type? Should they only run on the server type?

sibomots:
I would be OK with an API change that allows Task creation such that the Task is "shaded" to run on the rising (default) edge or on the falling edge of the ServerTickEvent.

As far as binding a Task to a Server Tick (default) vs CLIENT Tick, etc that could be done also. It just means altering the API. Perhaps the methods on Scheduler interface had type parameters analogous to template methods.

eg 

public Option<Task<?>> runTask(plugin, runnable, ....etc)

or adding more methods for each kind of Task.  Bloating the API vs typed Task objects.  

Either way I don't care, so long as it solves the problem efficiently.

sibomots:
Oops the editor mangled my example due to using chevrons and my phone cannot enter back ticks to put the code formatting.

PR #131, opened 2015-01-09
Implementation and Updating Submodules

piepie62:
I have updated the submodules and done my best at implementing new and older methods.

### DIFF
--- a/src/main/java/org/spongepowered/mod/entity/SpongeHorseColor.java
+++ b/src/main/java/org/spongepowered/mod/entity/SpongeHorseColor.java
@@ -25,11 +25,23 @@
 package org.spongepowered.mod.entity;
 
 import org.spongepowered.api.entity.living.meta.HorseColor;
+import org.spongepowered.api.service.persistence.DataSource;
+import org.spongepowered.api.service.persistence.data.DataContainer;
 
 public class SpongeHorseColor extends SpongeEntityMeta implements HorseColor {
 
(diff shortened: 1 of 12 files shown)


src/main/java/org/spongepowered/mod/entity/SpongeHorseColor.java
```diff
@@ -32,4 +34,14 @@ public SpongeHorseColor(int color, String name) {
         super(color, name);
     }
 
+    @Override
+    public DataContainer toContainer() {
+        return (DataContainer)this;
```
gabizou: 1) [`DataSerializable#toContainer()](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/service/persistence/DataSerializable.java#L38) is supposed to return a **NEW** DataContainer that represents the data of the serializable object, not casting the object as a DataContainer.

2) This cast will fail as none of the extending classes implement DataContainer.

piepie62: Okay, this is the one method that I am unsure about.

piepie62: Should I reduce it to a method stub in a new commit?

piepie62:
Okay, I didn't see PR 107, this is unneccesary. Closing.

PR #134, opened 2015-01-11
Fix max speed implementation in MixinEntityBoat

mproncace:
This PR corrects my implementation of the boat entity from #123. Previously, the custom max speed was not properly implemented.


bloodmc:
Merged thanks!

PR #136, opened 2015-01-12
Feature/queryscheduler

sibomots:
I left off the query methods for the scheduler.  This PR attempts to add the query methods to the scheduler.


sibomots:
Closed. Added feature of query methods to the scheduler.

PR #137, opened 2015-01-14
Add a vector helper to simplify vector conversions

simon816:
Makes it easier and neater
I also removed the switch statement for Direction <-> EnumFacing since it's no longer needed.

There's some use cases for this in #107 so if this gets merged first I'll update #107 to utilize VecHelper


bloodmc:
Looks good, I'll pull this tomorrow.

PR #138, opened 2015-01-16
Feature/async scheduler

sibomots:
This PR attempts to add the Asynchronous Scheduler to the code base.

The Tasks that are spun up by the Asynchronous Scheduler are based on wall-clock time and not relative to the in-game Tick as used in the other scheduler (SyncScheduler which implements Scheduler).

AsynchronousScheduler is an interface implemented in the Sponge core by the AsyncScheduler class.

The time granularity is down to the millisecond.

In all other respects the Tasks created by AsyncScheduler are going to behave just like Synchronous Tasks except in the timing of their start-up.

Additionally, the biggest change is that Tasks that are "asynchronous" cannot safely modify game data unless care is taken by the Plugin (not Sponge) to deal with those contingencies.  The JavaDoc preamble of the implementation class spells that out clearly.

There are several other ways to actually implement this Async version, and I've come across a few to consider. This is just one of many.

There are already some questions and comments so I expect to be making changes in the near future.


Mumfrey:
> The time granularity is down to the millisecond.

This would appear to not be true since it looks like you're using `System.currentTimeMillis()` which doesn't have guaranteed millisecond accuracy, for example on Windows the typical accuracy is ±15ms or so.

sibomots:
After some testing, I have some raw data.

See notes:  https://gist.github.com/sibomots/f5682cfd6d3f1b6529fe

sibomots:
Looks like the Async scheduler missed a step.    The ```startTask()```  method is going to be reworked so that the thread executor idiom is used instead to make each Task in ```AsyncScheduler``` truly async.   There'll be some thread pooling efficiency used.  Code coming soon.

sibomots:
I'd like to close this PR in favor of the changes in the updated branch feature/async-scheduler-2

zml2008:
you can close it yourself -- looks like you forgot to press the close&comment button

PR #140, opened 2015-01-18
Implement Egg

Aaron1011:
This PR implements `Egg`, and creates a foundation for other projectiles that extend `EntityThrowable` in NMS.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/entity/projectile/MixinEntityEgg.java
+++ b/src/main/java/org/spongepowered/mod/mixin/entity/projectile/MixinEntityEgg.java
@@ -1,0 +1,60 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/mod/mixin/entity/projectile/MixinEntityEgg.java
```diff
@@ -0,0 +1,61 @@
...
+@NonnullByDefault
+@Mixin(EntityEgg.class)
+public abstract class MixinEntityEgg extends EntityThrowable implements Egg {
+    public MixinEntityEgg(World worldIn) {
+        super(worldIn);
+    }
+
+    public double damageAmount;
+
+    @ModifyArg(method = "onImpact(Lnet/minecraft/util/MovingObjectPosition;)V", at =
```
Aaron1011: `onImpact` is an abstract method in `EntityThrowable`. Although it's always implemeted with a call to `attackEntityFrom`, it cannot be injected in `EntityThrowable`, as the bytecode is only present in a subclass.

src/main/java/org/spongepowered/mod/mixin/entity/projectile/MixinEntityEgg.java
```diff
@@ -0,0 +1,61 @@
...
+        @At(value = "INVOKE", target="Lnet/minecraft/entity/Entity;attackEntityFrom(Lnet/minecraft/util/DamageSource;F)Z"))
+    private float onAttackEntityFrom(float damage) {
+        System.out.println("Hai");
```
Zidane: Forgot a printout :p?

Aaron1011: Oops :D

src/main/java/org/spongepowered/mod/mixin/entity/projectile/MixinEntityThrowable.java
```diff
@@ -0,0 +1,84 @@
...
+
+    @Shadow
+    private String throwerName;
+
+    @Nullable
+    public ProjectileSource projectileSource;
+
+    @Override
+    public ProjectileSource getShooter() {
+        if (this.projectileSource != null && this.projectileSource instanceof ProjectileSource) {
```
simon816: No need for `instanceof` checks here and below.

Aaron1011: The `instanceof` check is actually needed for `getThrower`, as not all `EntityLivingBase` subclasses are `ProjectileSource`s.

PR #142, opened 2015-01-18
Implement Fireballs

AIOOB:


src/main/java/org/spongepowered/mod/mixin/entity/projectile/fireball/MixinEntityFireball.java
```diff
@@ -0,0 +1,81 @@
...
+        return this.projectileSource;
+    }
+
+    @Override
+    public void setShooter(ProjectileSource shooter) {
+        this.projectileSource = shooter;
+        if (shooter instanceof EntityLivingBase) {
+            this.shootingEntity = (EntityLivingBase) shooter;
+        } else {
+            this.shootingEntity = null;
```
gabizou: Seeing as I wrote the ProjectileSource, I was hoping that we could inject a new field variable for storing a reference to the `ProjectileSource` regardless whatever it is (dispenser etc).

src/main/java/org/spongepowered/mod/mixin/entity/projectile/fireball/MixinEntityFireball.java
```diff
@@ -0,0 +1,81 @@
...
+        super(worldIn);
+    }
+
+    @Override
+    public ProjectileSource getShooter() {
+        if (this.projectileSource == null || this.projectileSource != this.shootingEntity) {
+            if (this.shootingEntity != null) {
+                this.projectileSource = (ProjectileSource) this.shootingEntity;
+            } else {
+                this.projectileSource = new UnknownProjectileSource();
```
gabizou: This would be helpful to store in a field so that we can manage `BlockProjectileSource`s such as Dispensers to be used.

src/main/java/org/spongepowered/mod/mixin/entity/projectile/fireball/MixinEntityFireball.java
AIOOB: Isn't that what I have done?

gabizou: No, I'm talking about storing a new variable for the `ProjectileSource` in the event it's a `Dispenser`. That was the whole point of having the interface in the first place.

src/main/java/org/spongepowered/mod/mixin/MixinBootstrap.java
```diff
@@ -0,0 +1,66 @@
...
+import org.spongepowered.api.util.annotation.NonnullByDefault;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Overwrite;
+
+@NonnullByDefault
+@Mixin(targets = "net/minecraft/init/Bootstrap$8")
+public class MixinBootstrap {
+
+    @Overwrite
+    public ItemStack dispenseStack(IBlockSource source, ItemStack stack) {
```
Mumfrey: This overwrite is not necessary, and I've added a feature to `@Redirect` which makes it even easier to implement, see [this gist](https://gist.github.com/Mumfrey/b33797eb6ff3c991b433) which shows how to write an `@Redirect` handler to do what you need. I have also fixed the AP's handling of anonymous classes so if you update the Mixin submodule you should find that it now builds.

AIOOB:
In my workspace Mixin fails to compile as FakeElement and FakeType aren't implementing some inherited methods.

Mumfrey:
You're using the wrong JDK then.

PR #143, opened 2015-01-18
Implement server status ping API

Minecrell:
This implements the complete server status ping API from SpongePowered/SpongeAPI#367 and SpongePowered/SpongeAPI#331. It's implemented with a bunch of mixins to patch all the classes to make it work. :)

Patching `GameProfile` on the server is not possible right now because FML excludes the `com.mojang.*` classes from the transformer. Because of this, it's currently not usable in production on the dedicated server, as it requires MinecraftForge/FML#591 to run. I've added the modified `FMLServerTweaker` to the source so it runs in development environment.

I've chosen to rewrite the complete legacy ping handling because the version provided by Mojang is unstable and often disconnects clients because it doesn't get the complete data sent by the client due to fragmented TCP packets (see [this article in the Netty wiki](http://netty.io/wiki/user-guide-for-4.x.html#wiki-h3-11)). The new versions fixes this _for most cases_, there are still some cases where it doesn't work but these are nearly impossible to fix.

It's the first time I've used mixins so it would be great if I could get some feedback if I've used them correctly. :)


src/main/java/org/spongepowered/mod/mixin/entity/MixinEntity.java
```diff
@@ -332,4 +337,23 @@ public void setPersistent(boolean persistent) {
         // TODO
     }
 
+    @Override
+    public <T> Optional<T> getData(Class<T> dataClass) {
```
simon816: Technically this isn't related to the PR.

Minecrell: Yep, I've simply added these methods so we don't forget to implement them (because nothing complains that they're missing until you call them at runtime).

simon816:
I don't know what's wrong, but when attempting to connect with a client I get "Connection time out" error.
The server also appears offline in the server list.

Minecrell:
@simon816 Which client version did you use? It works fine for me.

simon816:
@Minecrell Vanilla 1.8.1.
I'll test again later in case it was something I had messed up

simon816:
@Minecrell It's most likely something my end since it works when running the server directly from the Sponge project in my eclipse setup, but from a test plugin project in the same workspace it doesn't.
I think it could be something wrong with my debug launch configuration, however it was working fine before but it's not a problem with this PR.

Minecrell:
@simon816 As mentioned in the description, it works only in development environment right now, not on the dedicated server, maybe this is the problem?

simon816:
@Minecrell No, it was completely a fault in my setup. I have a project 'Test Plugin' which I made the classpath inherit from 'Sponge', however the classpath entries were copied from 'Sponge' so it was referencing an older forge (the forge that was in use whenever I made the launch config). It was not a problem until this PR comes along and modifies a bunch of stuff for a later forge. It's all sorted now.

I guess that shows how important it is that sponge must be coupled with the same forge as it was built for.

PR #144, opened 2015-01-19
Plugin handling refactor

zml2008:
See commit comments, leave commentary


src/main/java/org/spongepowered/mod/mixin/fml/MixinInjectedModContainer.java
```diff
@@ -22,19 +22,20 @@
...
-public @interface PluginScoped {
+/**
+ * @author zml2008
+ */
+@Mixin(InjectedModContainer.class)
+public abstract class MixinInjectedModContainer implements ModContainer {
 
+    @Override
+    public boolean isImmutable() {
+        return !getModId().equals("SpongeAPIMod");
```
Kobata: Uhh... why not just do this in SpongeMod?

src/main/java/org/spongepowered/mod/plugin/SpongePluginContainer.java
```diff
@@ -40,45 +38,52 @@
...
+        } catch (NoSuchFieldException e) {
+            throw new ExceptionInInitializerError(e);
+        }
+        LANGUAGE_ADAPTER_FIELD.setAccessible(true);
+    }
 
     public SpongePluginContainer(String className, ModCandidate container, Map<String, Object> modDescriptor) {
-        // I suggest that you should be instantiating a proxy object, not the real plugin here.
-        super("org.spongepowered.mod.plugin.SpongePluginContainer$ProxyMod", container, modDescriptor);
+        super(className, container, modDescriptor);
```
Kobata: This is an error: We do /not/ want to be sending the plugin class to super as long as we're extending FMLModContainer.

Letting FMLModContainer know the actual class causes it to search that class for mod-related annotations.

src/main/java/org/spongepowered/mod/mixin/fml/MixinInjectedModContainer.java
Kobata: Looking at what this method does we really don't want to be changing it.

zml2008: If this is true, the mod won't appear in the mappings of mod container<->mod object, which we don't want. Do you have a better way to work around this issue?
InjectedModContainer hardcodes this as true (which wraps all coremod-specified ModContainers), we want Sponge to be accessible for lookups though.

src/main/java/org/spongepowered/mod/plugin/SpongePluginContainer.java
zml2008: could we not instead fix this by using the access transformer to make FMLModContainer.gatherAnnotations protected then overriding it in SpongePluginContainer to not scan for mod annotations?

src/main/java/org/spongepowered/mod/mixin/fml/MixinInjectedModContainer.java
Kobata: Loader.getIndexedModList(), we don't need to have an object instance.

zml2008: Sponge's API assumes we can go back and forth between container and instance of any mod or plugin.

src/main/java/org/spongepowered/mod/plugin/SpongePluginContainer.java
Kobata: There's more places than just there that check around the class it's given.

We need to either give it a fake class, or provide our own implementations of so much that we might as well stop extending FMLModContainer and implement the whole thing.

src/main/java/org/spongepowered/mod/mixin/fml/MixinInjectedModContainer.java
Kobata: Then implement it such that container = instance for anything without an FML instance (and/or have it bypass down to the wrapper container for Injected). There's going to be more things that violate that than just Sponge.

src/main/java/org/spongepowered/mod/plugin/SpongePluginContainer.java
zml2008: Can you link to the locations in FML where these checks would be different if we provided a fake proxy class?

src/main/java/org/spongepowered/mod/mixin/fml/MixinInjectedModContainer.java
zml2008: Looks like that's a possibility to do in Sponge's PluginManager. I'll do that instead.

src/main/java/org/spongepowered/mod/plugin/SpongePluginContainer.java
zml2008: I've had some fun with the access transformer to disable annotation processing methods, so none of the FML annotations are scanned.

bloodmc:
@zml2008 This PR almost ready to be pulled?

zml2008:
afaik it's ready to be pulled, unless more complaints are brought up

bloodmc:
@Kobata any other issues?

Kobata:
I do still think it's a bit... strange to be extending FMLModContainer and then basically ignoring a significant portion of what it does. That can be dealt with in a better manner later though.

src/main/java/org/spongepowered/mod/SpongeMod.java
```diff
@@ -134,10 +107,25 @@ public void onPreInit(FMLPreInitializationEvent e) {
     @Subscribe
     public void onInitialization(FMLInitializationEvent e) {
         this.registry.init();
+        try {
+            getGame().getServiceManager().setProvider(this, CommandService.class, new SimpleCommandService(getGame().getPluginManager()));
```
Aaron1011: It isn't possible to get a `PluginContainer` from a `SpongeMod` instance.

See https://github.com/SpongePowered/SpongeAPI/issues/426

zml2008: That's changed in this PR.

Aaron1011: Where? That's part of the API: `https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/service/SimpleServiceManager.java#L67`

zml2008: https://github.com/SpongePowered/Sponge/pull/144/files#diff-58e0014cf676b44128730ca938af8f0dR51
and 
https://github.com/SpongePowered/Sponge/pull/144/files#diff-877cd7506b19b24f60c4cdfd0180cf3cR60

Aaron1011: My bad, sorry.

src/main/resources/sponge_at.cfg
```diff
@@ -4,6 +4,12 @@ public net.minecraft.block.state.BlockState$StateImplementation
 public net.minecraft.network.handshake.client.C00Handshake field_149598_b # ip
 public net.minecraft.network.handshake.client.C00Handshake field_149599_c # port
 
+# Hook into FML machinery to customize mod loading more
+protected net.minecraftforge.fml.common.FMLModContainer languageAdapter
```
bloodmc: AT's don't work on FML/Forge classes.

PR #121, opened 2015-01-05
Add CommandBlock functionality

piepie62:
I have now fixed the oddities in my last pull request, except for the ones related to updating SpongeAPI and/or Mixin.

PR #145, opened 2015-01-19
[WIP] Implement Tile Entities.

me4502:
Note: Anyone can work on this, it's not just myself who is allowed to.


src/main/java/org/spongepowered/mod/mixin/block/data/MixinTileEntity.java
```diff
@@ -0,0 +1,47 @@
...
+@NonnullByDefault
+@Mixin(net.minecraft.tileentity.TileEntity.class)
+public abstract class MixinTileEntity implements TileEntity {
+
+    @Shadow
+    public abstract BlockPos getPos();
+
+    @Override
+    public BlockLoc getBlock() {
+        return (BlockLoc) getPos();
```
simon816: A MC `BlockPos` is not a Sponge `BlockLoc`

me4502: I see.. How does conversion happen?

simon816: Via the [`BlockWrapper`](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/wrapper/BlockWrapper.java) class.

src/main/java/org/spongepowered/mod/mixin/block/data/MixinTileEntitySign.java
```diff
@@ -0,0 +1,72 @@
...
+    }
+
+    public void sign$setLines(Message... lines) {
+        checkArgument(lines.length <= 4, "Only 4 lines can be entered on a sign!");
+        for(int i = 0; i < lines.length; i++)
+            signText[i] = ((SpongeMessage)lines[i]).getHandle();
+    }
+
+    public Message sign$getLine(int index) throws IndexOutOfBoundsException {
+        if(index > signText.length) throw new IndexOutOfBoundsException();
```
simon816: Guava has a [`checkElementIndex`](http://docs.guava-libraries.googlecode.com/git-history/release/javadoc/com/google/common/base/Preconditions.html#checkElementIndex(int, int)) which could be used here and below, this check currently fails if the index is negative.

gabizou:
I think that this can close #121 seeing as this provides a bit more functionality and is more feature complete.

me4502:
I believe the two PRs are actually doing different things. That is making CommandBlockLogic become a CommandSource, where is this allows interacting with the tile entity.

me4502:
@bloodmc Are you able to have a look over this PR? I'm aware some parts are missing, not sure how is the best way to finish those parts.

Deamon5550:
It occurs to me that the current layout of having `CommandBlock` extend `CommandSource` might actually be misleading. Instead there could be a getter `CommandSource getCommandSource()`. Especially since the way it will typically be implemented the tile entity is not actually a command source but is proxying into the actual command source so any object comparisons to check the command source would be false.

On the otherhand having the command block be the actual command source would allow a check along the lines of `source instanceof CommandBlock` to check the *type* of the source. This requires a change to how the command block is currently being mixed in.

But as it currently stands there are *two* command sources for each command block, the source of the tile entity and the actual source which is used for executing commands which is not actually exposed anywhere and does not allow object comparisons or instanceof checks by plugins.

me4502:
I haven't made the tile entity a CommandSource in this PR

Deamon5550:
See [the API for CommandBlock](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/block/data/CommandBlock.java#L34).

CommandBlock extends CommandSource, so yes your tile entitiy is a command source. In the perfect world this is correct because then to get the command source for a command block one only needs the tile entity. This also matches the pattern for other command sources (eg. the player is a command source).

However with your current implementation layout (and indeed minecraft's) the command source for a command block is separated away from the tile entity. So now you have two command sources, and the only accessible one is not actually the one to execute the commands, this is the problem.

So to reiterate there are two solutions:
A) Remove `CommandSource` from `CommandBlock` and add a `CommandSource getSource()` method.

B) (preferrable imo) Modify the TileEntity to be its own CommandSource in proper. This allows some additional functionality such as in a command handler `source instanceof CommandBlock` and similar useful things.

Aaron1011:
@me4502: I'm going to implement `Banner`, if that's all right with you.

me4502:
Sure, if you do it on your repo, message me and ill cherry pick it

AIOOB:
I have written some code implementing SkullType and NotePitch, if you would like it: https://github.com/AIOOB/Sponge/commit/2e61cabb6b1c191f422cc34d0d8e8e5d601979c2

me4502:
Thanks, added.

PR #152, opened 2015-01-24
Fix quit message being sent before the quit event

simon816:
~~Pretty bad hack TBH, will break easily if forge/MC change something.~~

~~This moves the call to `sendChatMsg` after the event has fired in order to get the message from the result of the event.~~

Much of the original PR contents has made it's way into SpongeCommon
This just completes the implementation by removing Forge's method call to fire the event due to Common already handling it.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/server/MixinServerConfigurationManager.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/server/MixinServerConfigurationManager.java
@@ -22,56 +22,33 @@
  * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
  * THE SOFTWARE.
  */
-package org.spongepowered.mod.mixin.core.fml.common.gameevent;
+package org.spongepowered.mod.mixin.core.server;
 
-import net.minecraftforge.fml.common.gameevent.PlayerEvent;
-import org.spongepowered.api.entity.player.Player;
-import org.spongepowered.api.event.entity.player.PlayerQuitEvent;
-import org.spongepowered.api.text.Text;
-import org.spongepowered.api.text.sink.MessageSink;
+import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.server.management.ServerConfigurationManager;
+import net.minecraftforge.fml.common.FMLCommonHandler;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
 
 @NonnullByDefault
-@Mixin(value = PlayerEvent.PlayerLoggedOutEvent.class, remap = false)
-public abstract class MixinPlayerLoggedOutEvent extends MixinPlayerEvent implements PlayerQuitEvent {
+@Mixin(ServerConfigurationManager.class)
+public abstract class MixinServerConfigurationManager {
 
-    private Text message;
-    private Text originalMessage;
-    private MessageSink messageSink;
-
-    @Override
-    public Text getMessage() {
-        return this.originalMessage;
+    /**
+     * @author Simon816
+     *
+     * Remove call to firePlayerLoggedOut because SpongeCommon's
+     * MixinNetHandlerPlayServer.onDisconnectPlayer fires the event already.
+     *
+     * NOTE: ANY call to playerLoggedOut will need to fire the
+     * PlayerLoggedOutEvent manually!
+     */
+    @Redirect(method = "playerLoggedOut", at = @At(value = "INVOKE",
+            target = "Lnet/minecraftforge/fml/common/FMLCommonHandler;firePlayerLoggedOut(Lnet/minecraft/entity/player/EntityPlayer;)V",
+            remap = false))
+    public void onFirePlayerLoggedOutCall(FMLCommonHandler thisCtx, EntityPlayer playerIn) {
+        // net.minecraftforge.fml.common.FMLCommonHandler.instance().firePlayerLoggedOut(playerIn);
     }
-
-    @Override
-    public Text getNewMessage() {
-        return this.message;
-    }
-
-    @Override
-    public void setNewMessage(Text quitMessage) {
-        if (this.originalMessage == null) {
-            // setNewMessage is always called before event fired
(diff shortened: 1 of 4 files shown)


simon816:
@modwizcode I could not think of a better way of doing it, I invite anyone else to find a neater solution but at least there is one right now though.
I updated the Mixin submodule as it needs to mix-in an inner class (`IntegratedServer$3`) which support for that was recently added to the Mixin repo.

modwizcode:
Alright this seems like a fine solution for now, if there's a better way we
can make it better then, but I cannot think of a solution now. I will have
to wait till I get home to fully evaluate it but for now :+1:

modwizcode:
@simon816 Is this not a forge issue? I understand needing a sponge patch, but I think the proper solution is a forge PR, since it's their event being fired at the wrong time.

simon816:
@modwizcode I don't think it's the event being fired at the wrong time, the event `PlayerLoggedOutEvent` is fired in `ServerConfigurationManager#playerLoggedOut` which does not handle the quit message (Since it's not needed for single player). It's best in forge's perspective to fire the event in `playerLoggedOut` but we need to be before that.
The login message can be delayed by forge but since forge doesn't utilize the message in the event, it's not needed.

src/main/java/org/spongepowered/mod/mixin/server/MixinServerConfigManager.java
```diff
@@ -0,0 +1,79 @@
...
+import org.spongepowered.mod.text.message.SpongeMessage;
+import org.spongepowered.mod.text.message.SpongeMessageTranslatable.SpongeMessageTranslatableBuilder;
+
+@Mixin(ServerConfigurationManager.class)
+public class MixinServerConfigManager {
+
+    private ChatComponentTranslation tmpJoinMessage;
+
+    // TODO Fix initializeConnectionToPlayer injection at runtime - it doesn't
+    // seem to know about FML's patches
```
simon816: Could someone take a look at this please? The [build fails](https://travis-ci.org/SpongePowered/Sponge/jobs/51739641#L522-L527) because it can't find the mappings for `initializeConnectionToPlayer`.
From some investigation the mappings don't seem to take into account FML's patch, [ServerConfigurationManager.java.patch](https://github.com/MinecraftForge/FML/blob/master/patches/minecraft/net/minecraft/server/management/ServerConfigurationManager.java.patch#L8)

me4502: @Mumfrey

Mumfrey: I'm not sure what we can do about this, this isn't actually the fault of your mixin or the AP. What's happening is that the patch is actually *altering the signature* of a public method, but the entry in the `srg` files is *not* updated with the corresponding information, so the `srg` still contains a mapping for the "old" (unpatched) version of the method.

This would be a problem for any code referencing that method (for example in a mod or mixin) because it means references won't get re-obfuscated and will thus be calling a non-existent method at production time. So basically the error is correct and we need to either:

1. Add the new reference the *srg* file so that it can be successfully reobfuscated, paging @AbrarSyed. A possible solution would be a companion file with the patches which specifies altered signatures which can be pumped into the srgs when they're generated.
2. Submit a PR to FML to fix this issue by doing whatever the patch is trying to to in a different way, since it could affect other code using that method as well.

Either way, it's not the mixin itself that needs to be fixed, and the AP error is telling us exactly what the issue is so the error is valid as well.

simon816: 145b9d673d4499bab486c9399d2eb84ea59b7e33 has fixed the issue

simon816:
Improved this slightly now that the `MixinServerConfigurationManager` exists and overwrites the `initializeConnectionToPlayer` method

simon816:
Updated to fix merge conflicts. Added comments on mixin injections.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinServerConfigurationManager.java
```diff
@@ -124,6 +129,15 @@
     @Shadow
     public abstract void playerLoggedIn(EntityPlayerMP playerIn);
 
+    /**
+     * @author bloodmc, Simon816
+     *
+     * TODO bloodmc needs to describe changes here
```
simon816: @bloodmc I have made a TODO for you to comment here :)

src/main/java/org/spongepowered/mod/mixin/core/server/MixinServerConfigurationManager.java
```diff
@@ -191,16 +205,6 @@ public void initializeConnectionToPlayer(NetworkManager netManager, EntityPlayer
...
-        ChatComponentTranslation chatcomponenttranslation;
-
-        if (!playerIn.getCommandSenderName().equalsIgnoreCase(s)) {
-            chatcomponenttranslation = new ChatComponentTranslation("multiplayer.player.joined.renamed", new Object[] {playerIn.getDisplayName(), s});
-        } else {
-            chatcomponenttranslation = new ChatComponentTranslation("multiplayer.player.joined", new Object[] {playerIn.getDisplayName()});
-        }
-
-        chatcomponenttranslation.getChatStyle().setColor(EnumChatFormatting.YELLOW);
-        this.sendChatMsg(chatcomponenttranslation);
```
Minecrell: If you keep everything except this here (comment this line out) you can minimize the diff, try to keep the diff between the original version and our overwrite as small as possible, that will simplify updating

Zidane:
While this can be pulled, Forge should get a PR to make this better (in the spirit of giving back).

simon816:
Something to consider is whether some of this should be moved to SpongeCommon, using the same technique as the join event (using `SpongeImplEventFactory`)

Zidane:
@simon816 

I'd move this to SpongeCommon (or at least all you can)

simon816:
I have moved as much as possible to SpongeCommon
https://github.com/SpongePowered/SpongeCommon/compare/feature/fire-quit-event
The rest of this PR is for removing Forge's method to fire the event and to handle single player.

simon816:
This is now ready to merge, as most of the implementation has moved to SpongeCommon this is just the icing on the cake.
This now simply removes Forge's method call to fire the quit event as it's already fired in SpongeCommon.
Also handles single player in the same way, except the quit message is never actually sent.

PR #153, opened 2015-01-25
Remove redundant casts

sirsavary:
After cloning Sponge I was browsing the code and came across a few redundant casts. I then used Intellij's inspection tool and came across a bunch more.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,3 +13,4 @@ matrix:
 cache:
   directories:
     - $HOME/.gradle/caches/
+    - $HOME/.gradle/wrapper/
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,3 +1,4 @@
+sudo: false
 language: java
 notifications:
   email: false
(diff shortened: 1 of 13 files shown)


sirsavary:
Sorta new to this Github thing, did not mean for my Travis commits to get lumped into this one.

Minecrell:
This is more something for the [`minor`](https://github.com/SpongePowered/Sponge/tree/minor) branch and I have fixed the redundant casts in https://github.com/SpongePowered/Sponge/commit/8cc1c9ee903a7d2ac74a6a73406313d53938f807 now.

For the Travis changes, most of this is in https://github.com/SpongePowered/Sponge/commit/aba88a2dbde9a6bb78513a9f183a0b241db31604 already. The dependency caching didn't really help a lot with our project because we don't use the Gradle wrapper right now and downloading the dependencies on every run was often even a little bit faster than packing and downloading the cache on every run.

PR #156, opened 2015-02-01
Added Tab Complete Functionality

verbeto:
Added functionality for tab complete.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/server/MixinNetHandlerPlayServer.java
+++ b/src/main/java/org/spongepowered/mod/mixin/server/MixinNetHandlerPlayServer.java
@@ -57,12 +57,9 @@ import java.util.List;
 @Mixin(NetHandlerPlayServer.class)
 public abstract class MixinNetHandlerPlayServer implements PlayerConnection, INetHandlerPlayServer {
 
-    @Shadow
-    public NetworkManager netManager;
-    @Shadow
-    public EntityPlayerMP playerEntity;
-    @Shadow
-    private MinecraftServer serverController;
+    @Shadow public NetworkManager netManager;
+    @Shadow public EntityPlayerMP playerEntity;
+    @Shadow private MinecraftServer serverController;
 
     @Shadow
     public abstract void sendPacket(final Packet packetIn);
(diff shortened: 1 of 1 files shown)


simon816:
Please follow the style guide. No tabs, 4 spaces per indentation level.

Also, remove 'Update MixinNetHandlerPlayServer.java' from the commit message, it doesn't mean anything.

verbeto:
Fixed.

src/main/java/org/spongepowered/mod/mixin/server/MixinNetHandlerPlayServer.java
```diff
@@ -79,4 +97,57 @@ public void sendCustomPayload(Object plugin, String channel, byte[] data) {
...
+            arraylist.add(value);
+        }
+        this.playerEntity.playerNetServerHandler.sendPacket(new S3APacketTabComplete((String[]) arraylist.toArray(new String[arraylist.size()])));
+
+    }
+
+    @SuppressWarnings({ "rawtypes", "unchecked" })
+    private List<String> processTab(ICommandSender sender, String string) {
+
+        ArrayList arraylist = Lists.newArrayList();
```
kenzierocks: This doesn't need to be a rawtype.

verbeto:
Warnings fixed also.

kenzierocks:
You've got a few cast warnings.

simon816:
This can be done neater with `@Inject` instead of copying the entire method.

You don't need to implement `INetHandlerPlayServer` as that is already an interface on the target class and you're missing an `@Overwrite` (Though should be `@Inject` ideally)

simon816:
@verbeto Are you able to fix the above issues? Currently this will not work in an obfuscated environment.

verbeto:
Apologies for the delay, the implementation has been rewritten and now uses the Overwrite annotation.    Unfortunately, as per the comment of the request, Inject cannot be used as more than one tab complete packet would result in only the final packet being displayed, and besides; implementing it this way gives us more control.

simon816:
I think a better place for this would be in `MinecraftServer.func_180506_a` (getTabCompletions) or even `CommandHandler.getTabCompletionOptions`

verbeto:
Well, I'll leave it like this- personally, I think separating vanilla code and sponge code as much as possible is best but, if need be, I can move the code into one of the methods you listed.

zml2008:
This has been done with PR #215's refactoring

PR #157, opened 2015-02-03
Add code for getMinecraftVersion()

sirsavary:
With the help of Sibomots I wrote some code so getMinecraftVersion() in SpongeGame would work. However, we're not sure if the singleton approach is the best.

~~I also fixed "gradle javadoc" from throwing errors but there is a file that needs to be fixed. I removed a "@see" annotation thinking it was invalid when really the error was caused by the annotation being immediately followed by a "@link" annotation.~~

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeGame.java
+++ b/src/main/java/org/spongepowered/mod/SpongeGame.java
@@ -25,6 +25,7 @@
 package org.spongepowered.mod;
 
 import com.google.common.base.Optional;
+import net.minecraft.server.MinecraftServer;
 import net.minecraftforge.fml.common.FMLCommonHandler;
 import org.spongepowered.api.Game;
 import org.spongepowered.api.GameRegistry;
@@ -97,7 +98,7 @@ public final class SpongeGame implements Game {
 
     @Override
     public MinecraftVersion getMinecraftVersion() {
-        throw new UnsupportedOperationException(); // TODO
+        return (MinecraftVersion) MinecraftServer.getServer().getServerStatusResponse().getProtocolVersionInfo();
     }
 
     @Override
--- a/src/main/java/org/spongepowered/mod/SpongeMinecraftVersion.java
+++ b/src/main/java/org/spongepowered/mod/SpongeMinecraftVersion.java
@@ -25,6 +25,7 @@
 package org.spongepowered.mod;
 
 import com.google.common.base.MoreObjects;
+import net.minecraft.network.ServerStatusResponse;
 import org.spongepowered.api.MinecraftVersion;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
 
@@ -37,6 +38,11 @@ public class SpongeMinecraftVersion implements ProtocolMinecraftVersion {
     public SpongeMinecraftVersion(String name, int protocol) {
         this.name = name;
         this.protocol = protocol;
+    }
+
+    public SpongeMinecraftVersion(ServerStatusResponse.MinecraftProtocolVersionIdentifier identifier) {
+        this.name = identifier.getName();
+        this.protocol = identifier.getProtocol();
     }
 
     @Override


Walkman100:
* @sibomots :P

sibomots:
I'm not sure that this implementation is in the right direction.

A few issues:

The Legacy flag is false, and nothing in the code will change it.  Please review the API definition of the method.

I don't think that a singleton is really necessary.  There are other classes involved with the Minecraft version discovery in the package.

The compareTo methid is not correct.  You're comparing references, while the intent is to compare Minecraft versions and Legacy values

sirsavary:
I don't really like the singleton solution as it clogs up SpongeGame and isn't super intuitive. I originally set the Legacy flag to false becuase I was told only one class extended MinecraftVersion, but that information was incorrect. I think the best method would be to return either an existing SpongeMinecraftVersion or a SpongeLegacyMinecraftVersion.

kenzierocks:
The javadoc stuff should probably not be in this PR, it's not part of the version stuff.

src/main/java/org/spongepowered/mod/service/scheduler/SyncScheduler.java
```diff
@@ -41,7 +41,7 @@
 /**
  * <p>Internal implementation of the SynchronousScheduler interface.</p>
  *
- * @see {@link org.spongepowered.api.service.scheduler.SynchronousScheduler}
+ * See {@link org.spongepowered.api.service.scheduler.SynchronousScheduler}
```
kenzierocks: Not the right way to change this, it should be:
```
@see org.spongepowered.api.service.scheduler.SynchronousScheduler
```

sirsavary: I'll revert it and throw it in another PR, thanks for providing the correct syntax.

src/main/java/org/spongepowered/mod/SpongeGame.java
```diff
@@ -97,7 +101,24 @@ public String getImplementationVersion() {
...
+	    MinecraftVersion mcvToReturn;
+
+	    // Get the MC version and protocol version from the MC server, then parse it so we can use it earlier
+	    ServerStatusResponse.MinecraftProtocolVersionIdentifier protocolVersionInfo = MinecraftServer.getServer().getServerStatusResponse().getProtocolVersionInfo();
+	    String name = protocolVersionInfo.getName();
+	    int protocolVersion = protocolVersionInfo.getProtocol();
+	    double numberVersion = Double.valueOf(name);
+
+	    // If the version is less than 1.7 it is considered legacy and must be handled differently
+	    if (numberVersion < 1.7) {
```
bloodmc: Our implementation will never support legacy so there is no need to check for 1.7 and under.

src/main/java/org/spongepowered/mod/SpongeGame.java
```diff
@@ -97,7 +101,24 @@ public String getImplementationVersion() {
 
     @Override
     public MinecraftVersion getMinecraftVersion() {
-        throw new UnsupportedOperationException(); // TODO
+	    MinecraftVersion mcvToReturn;
+
+	    // Get the MC version and protocol version from the MC server, then parse it so we can use it earlier
+	    ServerStatusResponse.MinecraftProtocolVersionIdentifier protocolVersionInfo = MinecraftServer.getServer().getServerStatusResponse().getProtocolVersionInfo();
```
bloodmc: These lines can be removed. Just add a constructor in SpongeMinecraftVersion that accepts MinecraftProtocolVersionIdentifier so you can do the following

``` java
serverVersion = new SpongeMinecraftVersion(MinecraftServer.getServer().getServerStatusResponse().getProtocolVersionInfo());
``` java

**Comment by Minecrell** [comment]
Keep in mind `ServerStatusResponse.MinecraftProtocolVersionIdentifier` actually already implements `MinecraftVersion`, there is no real need to create another `SpongeMinecraftVersion` every time.

**Comment by sibomots** [comment]
What were the test results?  Some sample code would be nice.

But regardless, there's some issues with this PR, it doesn't really fix the underlying problem.

There are two methods that need to be fixed.

Was:

```java
    @Override
    public boolean equals(Object o) {
        if (this == o) {
            return true;
        }
        if (!(o instanceof SpongeMinecraftVersion)) {
            return false;
        }

        SpongeMinecraftVersion that = (SpongeMinecraftVersion) o;
        return this.protocol == that.protocol;

    }
```

is:

```java
    @Override
    public boolean equals(Object obj) {
        if ( this == obj ) {
            return true;
        }
        if  ( ! ( obj instanceof SpongeMinecraftVersion ) ) {
            return false;
        }

        SpongeMinecraftVersion that = obj;
        return  ( this.protocol == that.protocol  && this.name.equals(that.getName() ) ) );

    }
```


```java
    public static int compare(ProtocolMinecraftVersion version, MinecraftVersion to) {
        if  (version.equals( to ) ) {
            return 0
        } else if (  to.isLegacy()  ) {
            return 1;
        } else {
            boolean versionOrder =   version.getProtocol()  < ( ( ProtocolMinecraftVersion ) to ).getProtocol();
            
            return versionOrder ? -1 : 1;
        }
    }
```

src/main/java/org/spongepowered/mod/SpongeMinecraftVersion.java
```diff
@@ -39,6 +40,11 @@ public SpongeMinecraftVersion(String name, int protocol) {
         this.protocol = protocol;
     }
 
+    public SpongeMinecraftVersion(ServerStatusResponse.MinecraftProtocolVersionIdentifier identifier) {
```
simon816: This constructor is not needed since it's not used.

sirsavary: Blood told me to throw it in there, he said it should be there. I can remove it though and let someone else add it.

PR #159, opened 2015-02-06
Places imports in order to satisfy checkStyle

sibomots:
This PR will address one section of the compiler checkstyle warnings.

It re-orders imports according to the checkstyle predicates.

Don't use "IDE" formatting to handle it.  

The order for new code must be:

import static

import org.spongepowered.*

import third party

import java*

Each of the four sections, the imports are alphabetized.


kenzierocks:
This seems almost impossible to follow, if we can't just IDE format it. Do you think there is any tool we could integrate to order them automagically?

PR #160, opened 2015-02-06
Import order redux

sibomots:
This PR will address one section of the compiler checkstyle warnings.

It re-orders imports according to the checkstyle predicates.

Don't use "IDE" formatting to handle it.

The order for new code must be:

import static

import org.spongepowered.*

import third party

import java*

Each of the four sections, the imports are alphabetized.

### DIFF
--- a/src/main/java/org/spongepowered/api/text/action/SpongeTextActionFactory.java
+++ b/src/main/java/org/spongepowered/api/text/action/SpongeTextActionFactory.java
@@ -24,8 +24,6 @@
  */
 package org.spongepowered.api.text.action;
 
-import java.net.URL;
-
 import org.spongepowered.api.entity.Entity;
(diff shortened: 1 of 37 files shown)


simon816:
There should not be a separate section for `org.spongepowered`, see [Sponge Contributing Guidelines](https://docs.spongepowered.org/en/latest/devs/guidelines/)
There is a [`sponge.importorder`](https://github.com/SpongePowered/SpongeAPI/blob/master/extra/sponge.importorder) file for reference.
A change to the guidelines should be made before this if there should be something different.

Minecrell:
The import order is/was fixed in the `minor` branch already. For the few import warnings still coming up with it we have a custom checkstyle version that fixes that. And I agree with @simon816 here, `org.spongepowered` is not an extra group and we have no exception for it in our contribution guidelines.

PR #162, opened 2015-02-07
Fix CommandBlock sources fixes #161

### LINKED ISSUE #161: CommandBlock CommandSource
Currently TileEntityCommandBlock$1 which is the class implementing ICommandSender and is the actual object passed to the command handlers doesn't implement sponge's CommandSource which causes the server to crash whenever a commandblock is activated. However the fix is not quite as simple as it may seem at first glance. There are three requirements for a correct solution in descending importance (only the first is absolutely required):

- **CommandSource**: The object passed to the internal command handler must implement CommandSource in order to stop the crash from occurring.
- **instanceof**: The class passed to the command handler should also implement CommandBlock in order to allow an instanceof check in a plugin's CommandCallable to see if the command is originating from a command block.
- **equals**: The class passed to the command handler should be the actual instance of the command block, to allow an equals check against a specific tile entity in a plugin's CommandCallable.

There are three possible solutions that I currently see.

**Solution A: Inner class implements CommandSource**
This is the simplest solution and does stop the crash from occurring. However this does not solve the instanceof or equals requirements. No other immediate drawbacks that I can see.

**Solution B: Inner class implements CommandBlock**
By implementing commandblock you are implicitly implementing CommandSource thus solving the first issue. You are also solving the instanceof requirement as the source passed to a plugin's CommandCallable will be a class implementing CommandBlock. However, the object passed to the command handler will still be a different object to the actual command block tile entity sourcing the command so it does not cover the equals requirement. Additionally this is a bit of a janky solution as the inner class would have to proxy all extraneous CommandBlock methods back to the parent tile entity.

**Solution C: TileEntityCommandBlock implements ICommandSender and bypasses the inner class**
By having the actual tile entity implement ICommandSender and CommandBlock we can pass the actual object of the tile entity to the command handler thus solving all three requirements. However this is quite an invasive change as any other forge mods expecting the origin of the commands from the commandblock to be the inner class will fail.

An alternative would be if it was possible to send one object to the minecraft/forge command handler (the inner class) and another to the sponge command handler (the tile entity). However I'm not sure of the feasibility of this within the current system.

As you can see there is no perfect solution so any input would be appreciated.

Deamon5550:
This PR fixes #161 by mixing the new `CommandBlockSource` into minecraft's `CommandBlockLogic`. This ensures that A) the object passed to the command handler is comparable with sponge's own command handler and B) the CommandSource sent to a plugin's CommandCallable can be checked using `source instanceof CommandBlockSource` to determine its *type* so to speak.

This single mixin will affect both `TileEntityCommandBlock` and `EntityMinecartCommandBlock` as both use `CommandBlockLogic` as the handler for their command internals.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/command/MixinCommandBlockLogic.java
+++ b/src/main/java/org/spongepowered/mod/mixin/command/MixinCommandBlockLogic.java
@@ -1,0 +1,76 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/mod/mixin/command/MixinCommandBlockLogic.java
```diff
@@ -0,0 +1,76 @@
...
+    @Override
+    public void sendMessage(Iterable<Message> messages) {
+        for (Message msg : messages) {
+            this.addChatMessage(((SpongeMessage<?>) msg).getHandle());
+        }
+    }
+
+    @Override
+    public Location getLocation() {
+        return new Location((Extent) getEntityWorld(), new Vector3d(getPositionVector().xCoord, getPositionVector().yCoord,
```
simon816: This could be added to [`VecHelper`](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/util/VecHelper.java) for convenience

Deamon5550: Yeah, there are a bunch of useful vector conversions that could be added to VecHelper, like something for Vec3/Vector3d as well.

PR #165, opened 2015-02-08
Implement more projectiles

AIOOB:
I have had to leave MixinBootstrap commented out as running `gradle build` still gives an error, but it runs perfectly fine in my IDE.


src/main/java/org/spongepowered/mod/mixin/entity/projectile/MixinEntityEnderEye.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+
+    @Override
+    public ProjectileSource getShooter() {
+        return this.projectileSource;
+    }
+
+    @Override
+    public void setShooter(ProjectileSource shooter) {
+        this.projectileSource = shooter;
```
AzureusNation: Only one problem with this. If the server was to crash and restart and then once restarted a plugin tries to get the shooter of that projectile, its just going to return Unknown as it isn't being saved anywhere.

simon816: Shouldn't be too much of a problem, just need to inject a bit of code into the save/load NBT methods. Need to make `ProjectileSource` serializable to NBT then it should be fine.

AzureusNation: @simon816 Good point, I like this guy :P

gabizou: Which is really easy to do. I believe there was a more persistent way of saving custom data that @bloodmc was pointing out. I believe most of these custom data saving features will have to be done en-masse for that.

simon816: I'm currently fixing this and a bunch of other places where entity mixin data isn't being persisted (See branch [persistent-entity-fix](https://github.com/simon816/Sponge/compare/persistent-entity-fix)
In vanilla, the shooter is not persisted, should I make it persist? I can see why not since most projectiles have a short lifespan.

src/main/java/org/spongepowered/mod/mixin/block/MixinBehaviorProjectileDispense.java
```diff
@@ -0,0 +1,50 @@
...
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Redirect;
+
+@NonnullByDefault
+@Mixin(net.minecraft.dispenser.BehaviorProjectileDispense.class)
+public class MixinBehaviorProjectileDispense extends BehaviorDefaultDispenseItem {
+
+    @Redirect(method = "dispenseStack", at = @At(value = "INVOKE", target = "Lnet/minecraft/world/World;spawnEntityInWorld(Lnet/minecraft/entity/Entity;)Z"))
+    public boolean onSpawnEntityInWorld(World world, Entity entity, IBlockSource source, ItemStack stack) {
+        ((Projectile) entity).setShooter((ProjectileSource) source.getBlockTileEntity());
```
simon816: Should definitely check the entity is an instance of `Projectile`. (example: Modded entities may not be)
```java
TileEntity tileEntity = source.getBlockTileEntity();
if (entity instanceof Projectile && tileEntity instanceof ProjectileSource) {
    ((Projectile) entity).setShooter((ProjectileSource) tileEntity);
}
return world.spawnEntityInWorld(entity);
```

AIOOB:
I have removed the NBT due to #173, depending on which one gets pulled first one of us will update our own to fix it.

PR #166, opened 2015-02-09
Implement Item

Aaron1011:
This implements `Item`.

Because the Item interface deliberately doesn't expose the underlying magic numbers, it's necessary to keep track of whether the age and pickup time were set through the api or not. This allows vanilla behavior to be preserved in the case where the magic number is set some other way (e.g. a forge mod, **/summon**, etc)

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/entity/MixinEntityItem.java
+++ b/src/main/java/org/spongepowered/mod/mixin/entity/MixinEntityItem.java
@@ -1,0 +1,172 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/mod/mixin/entity/MixinEntityItem.java
```diff
@@ -0,0 +1,170 @@
...
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
+
+import javax.annotation.Nullable;
+
+@NonnullByDefault
+@Mixin(EntityItem.class)
+public abstract class MixinEntityItem extends Entity implements Item {
+
+    private static final int MAGIC_INFINITE_PICKUP_DELAY = 32767;
```
kenzierocks: Possibly hardcode these to the Short constants?

kenzierocks:
I like it.

src/main/java/org/spongepowered/mod/mixin/entity/MixinEntityItem.java
Zidane: Agreed with @kenzierocks

Zidane:
Fix what @kenzierocks and myself commented on above @Aaron1011 and conflicts.

modwizcode:
@Zidane this looks good now, can it be merged?

src/main/resources/mixins.sponge.json
```diff
@@ -45,6 +45,7 @@
         "entity.hanging.MixinEntityLeashKnot",
         "entity.hanging.art.MixinEnumArt",
         "entity.living.MixinEntityAgeable",
+	"entity.living.MixinEntityItem",
```
gabizou: Space

src/main/java/org/spongepowered/mod/mixin/entity/MixinEntityItem.java
```diff
@@ -0,0 +1,170 @@
...
+        this.delayBeforeCanPickup = MAGIC_INFINITE_PICKUP_DELAY;
+        this.pluginPickupSet = true;
+        this.infinitePickupDelay = true;
+    }
+
+    @Override
+    public int getDespawnTime() {
+        if (this.age == MAGIC_INFINITE_DESPAWN_TIME) {
+            if ((this.pluginDespawnSet && this.infiniteDespawnDelay) || !this.pluginDespawnSet) {
+                return -1;
```
Mumfrey: Given that you have consts for the other magic numbers, would it not have a pleasing symmetry to also declare these two occurrences of -1 as such?

src/main/resources/mixins.sponge.json
```diff
@@ -45,6 +45,7 @@
         "entity.hanging.MixinEntityLeashKnot",
         "entity.hanging.art.MixinEnumArt",
         "entity.living.MixinEntityAgeable",
+        "entity.living.MixinEntityItem",
```
simon816: Wrong package, should be `entity.MixinEntityItem`

PR #167, opened 2015-02-09
Fill out Wolf implementation

Aaron1011:
I've also implemented `EntityTameable`, and added the necessary `getName` to `EntityPlayer` so that it can be used as a `Tamer`

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityTameable.java
+++ b/src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityTameable.java
@@ -1,0 +1,134 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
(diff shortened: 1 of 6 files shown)


Minecrell:
If you add `getName()` to `EntityPlayer`, you should be able to remove `getName()` from EntityPlayerMP.

src/main/java/org/spongepowered/mod/mixin/entity/player/MixinEntityPlayer.java
```diff
@@ -71,4 +75,8 @@ public MixinEntityPlayer(World worldIn) {
         return this.openContainer == this.inventoryContainer;
     }
 
+    public String human$getName() {
+        return this.getName();
```
Minecrell: Like https://github.com/SpongePowered/Sponge/commit/07d73f4b74f7851bf23b81bd01a4998a6d9df8fc#commitcomment-9648317, this will cause an infinite loop in the deobfuscated environment because `this.getName()` and `human$getName()` will be the same. You need to implement it like https://github.com/SpongePowered/Sponge/commit/07154247ec9fe21e81117a43a2326de0e20c3bf7

Aaron1011: Oops, I thought I had added that. I'll fix it.

Zidane:
@Aaron1011 

Fix conflicts.

src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityTameable.java
```diff
@@ -0,0 +1,134 @@
...
+    }
+
+    @Overwrite
+    public Entity getOwner() {
+        return this.func_180492_cm();
+    }
+
+    public void sittable$setOwner(Tamer tamer) {
+        if (tamer instanceof EntityPlayer) {
+            this.func_152115_b(((EntityPlayer) tamer).getName());
```
simon816: Should be a UUID string not a name. See the usage of `func_152115_b` in vanilla

src/main/java/org/spongepowered/mod/mixin/entity/player/MixinEntityPlayer.java
simon816: A new way of fixing this error has just been pushed to the Mixin repo, you can use an [`@Intrinsic`](https://github.com/SpongePowered/Mixin/blob/master/src/main/java/org/spongepowered/asm/mixin/Intrinsic.java) annotation.

Mumfrey: Actually just for the record, the way @Aaron1011 has done this is exactly correct. By placing the `@Overwrite` with the original code *below* the interface accessor, it will *not* cause the infinite loop. However @simon816 is absolutely correct, this can now be solved with [`@Intrinsic`](https://github.com/SpongePowered/Mixin/blob/master/src/main/java/org/spongepowered/asm/mixin/Intrinsic.java) because the purpose of `Intrinsic` is precisely to avoid this kind of unnecessary duplication.

So basically, this code is quite correct - per my examples - but using `Intrinsic` will make it better.

Aaron1011: I'm about to push code that changes it - I'm just debugging some other issues first.

src/main/java/org/spongepowered/mod/mixin/entity/living/MixinEntityTameable.java
```diff
@@ -0,0 +1,136 @@
...
+    @Shadow
+    public abstract void func_175544_ck();
+
+    @Shadow
+    public abstract EntityLivingBase func_180492_cm();
+
+    @Shadow
+    public abstract void func_152115_b(String name);
+
+    private Tamer tamer;
```
simon816: This field will not persist over server restarts, see https://github.com/SpongePowered/Sponge/pull/173 for how to fix the issue.

Aaron1011:
This is blocked on (more of) the Data API being implemented.

PR #172, opened 2015-02-15
Implement PlayerDropItemEvent

Aaron1011:
Due to the fact that `ItemTossEvent` extends `ItemEvent` and not `PlayerEvent`, it's necessary to implement `getPlayer`, `getHuman`, and `getLiving`, as they won't be mixed into a superclass.


bloodmc:
Rename Mixin class to MixinEventPlayerDropItem and store it in the event.player package.

Aaron1011:
Closed in https://github.com/SpongePowered/Sponge/commit/72b8f1b856ff773e7d57d498992dc62189c5f34c

PR #169, opened 2015-02-14
Add EnderPearl damage methods back

Aaron1011:
Now that `EnderPearl` is a `DamagingProjectile`, it once again has damage getters and setters.

The code in `EnderPearl` is identical to what it was before the methods were removed - I've just updated some other files affected by the commits pulled in by the submodule update.

PR #173, opened 2015-02-15
Fix some entity mixins not storing their state

simon816:
Initially discovered in https://github.com/SpongePowered/Sponge/pull/165#discussion_r24569829.

Entity mixins that define additional data do not save this data to the disk since vanilla is not aware it.
I have gone through every entity mixin to find any data that will not persist.
It is unlikely for a lightning bolt to need to be persistent but I have added it anyway.

The following pull requests may have some data that should be made persistent.
~~#126~~, ~~#165~~, ~~#166~~, #167, ~~#169~~ 
I can either wait for the PRs to be pulled, then update this one to fix the issues. Or pull this and I can suggest changes needed for the other PRs.

### DIFF
--- a/src/main/java/org/spongepowered/mod/entity/projectile/ProjectileSourceSerializer.java
+++ b/src/main/java/org/spongepowered/mod/entity/projectile/ProjectileSourceSerializer.java
@@ -1,0 +1,91 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered.org <http://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+
+package org.spongepowered.mod.entity.projectile;
+
+import net.minecraft.nbt.NBTBase;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.nbt.NBTTagLong;
+import net.minecraft.nbt.NBTTagString;
+import net.minecraft.tileentity.TileEntity;
+import net.minecraft.util.BlockPos;
+import net.minecraft.world.World;
+import org.spongepowered.api.entity.Entity;
+import org.spongepowered.api.entity.projectile.Projectile;
+import org.spongepowered.api.entity.projectile.source.BlockProjectileSource;
+import org.spongepowered.api.entity.projectile.source.ProjectileSource;
+import org.spongepowered.api.entity.projectile.source.UnknownProjectileSource;
+import org.spongepowered.mod.util.VecHelper;
+
+import java.util.UUID;
+
+public class ProjectileSourceSerializer {
+
+    // TODO Revisit when persistent containers are implemented.
+    // Note: ProjectileSource itself does not extend DataContainer
+
+    public static NBTBase toNbt(ProjectileSource projectileSource) {
+        if (projectileSource instanceof Entity) {
+            return new NBTTagString(((Entity) projectileSource).getUniqueId().toString());
+        }
+        if (projectileSource instanceof BlockProjectileSource) {
+            return new NBTTagLong(VecHelper.toBlockPos(((BlockProjectileSource) projectileSource).getBlock().getPosition()).toLong());
+        }
+        return null;
+    }
+
+    public static ProjectileSource fromNbt(World worldObj, NBTBase tag) {
+        if (tag instanceof NBTTagString) {
+            Entity entity =
+                    ((org.spongepowered.api.world.World) worldObj).getEntity(UUID.fromString(((NBTTagString) tag).getString())).orNull();
+            if (entity instanceof ProjectileSource) {
+                return (ProjectileSource) entity;
+            }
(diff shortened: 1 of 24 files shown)


Mumfrey:
This looks good. Are the additional data supposed to be stored in the NBT though? The problem with storing additional data in the NBT is that if the world is loaded in a non-sponge implementation and the entity is read then written then the additional data are lost.

Would it instead be worthwhile to leverage the serialisation API instead and store all the additional data in a separate folder tree? Decoupling the custom data would preserve them across loads where Sponge isn't present.

In the days of CraftBukkit, the chances of this happening were pretty slim unless the server owner specifically launched vanilla (like when a new minecraft came out and a server owner wanted to skip ahead and use vanilla until the new CB came out), but since Sponge is a mod there are other circumstances that could cause it to not be loaded and the user to not know about it until it's too late and the damage is already done:

For example, the user is running other coremods and one of them interferes with the loading of Sponge, Sponge doesn't load and the user merrily loads up their world and kills all the extra data for entities in loaded chunks. Or a user running Sponge client-side upgrades their forge installation to a version not supported by the version of sponge they have installed and the same thing happens.

Worth thinking about. Since we don't control the vertical like CraftBukkit did, we can't make the same assumptions unfortunately.

AIOOB:
This is our conversation with @bloodmc: [Chat log](http://sbnc.khobbits.co.uk/log/logs/old/spongedev_%5B2015-02-12%5D.htm)
Although we didn't really consider Sponge not being loaded, it has some interesting points in there.

Mumfrey:
@AIOOB it looks from that conversation like @simon816 suggested the sensible idea of putting the data under `ForgeData` so that the data aren't lost at least if Forge is present. Since this is a Forge-based implementation that would make sense and would address the two issues I mentioned above (if Forge does indeed preserve keys in that tag even if a specific mod isn't present) but then this seems to counteract it

> <AIOOB> simon816: Is it really neccesary? It is probably a good thing to be less reliant on forge maintaining the same structure.
> <simon816> Doesn't matter really, just thought that method could be used
> <@blood|work> simon816: I prefer it to be seperate NBT
> <@blood|work> not within Forge
> <simon816> OK
> <@blood|work> SpongeData sounds good

which opens the original problem right back up again, but without a technical reason but seemingly for reasons of neatness, which doesn't seem like a good reason to abandon a more reliable solution.

Reading up on Forge's mechanisms, the standard suggestion seems to be calling `Entity::getEntityData()`` which returns the underlying `ForgeData` compound tag. This seems to address @bloodmc's concern of *"forge maintaining the same structure* since having a reliance on (what seems to be) a fairly core method for serialisation has a very different flavour to relying on an implementation detail, so if Forge does change the underlying storage in the future it won't be a breaking change.

Calling `getEntityData` and then storing everything in a `SpongeData` tag within it seems like it would fix the sponge-possibly-not-being-loaded issue, provided Forge is always in use (which is a more reasonable assumption).

bloodmc:
@Mumfrey brings up a good point about loading in a non-sponge environment. As Sponge is technically just another mod, it should be placed with all other mod data.
@simon816 Please store all Sponge data in a compound labeled "SpongeData" within the ForgeData compound

simon816:
Done.

I think that if Sponge didn't load successfully, it should force stop the server in order to avoid things going wrong and make sure plugins don't run. Or should block the server from loading until the user decides whether it should continue (Just like FML has the `/fml confirm`)

src/main/java/org/spongepowered/mod/entity/projectile/ProjectileSourceSerializer.java
```diff
@@ -0,0 +1,88 @@
...
+import java.util.UUID;
+
+public class ProjectileSourceSerializer {
+
+    public static NBTBase toNbt(ProjectileSource projectileSource) {
+        if (projectileSource instanceof Entity) {
+            return new NBTTagString(((Entity) projectileSource).getUniqueId().toString());
+        }
+        if (projectileSource instanceof BlockProjectileSource) {
+            return new NBTTagLong(VecHelper.toBlockPos(((BlockProjectileSource) projectileSource).getBlock().getPosition()).toLong());
```
gabizou: Just out of curiosity, wouldn't this be better handled with `ProjectileSource#toContainer()` of which we can serialize the container internally? Regardless how it's actually serialized?

Perhaps a `TODO` for future persistence purposes.

src/main/java/org/spongepowered/mod/mixin/entity/projectile/MixinEntityFishHook.java
```diff
@@ -206,4 +209,18 @@ public void setFishingRodItemStack(ItemStack fishingRod) {
         this.fishingRod = fishingRod;
     }
 
+    @Override
+    public void readFromNbt(NBTTagCompound compound) {
+        // TODO super.readFromNbt(compound);
+        if (compound.hasKey("damageAmount")) {
```
kenzierocks: This is never written.

simon816: Oops, must have accidentally removed the write when playing about with `ProjectileSourceSerializer`. I'll fix this

src/main/java/org/spongepowered/mod/entity/projectile/ProjectileSourceSerializer.java
simon816: Good thought. I'll add a TODO

Mumfrey:
> I think that if Sponge didn't load successfully, it should force stop the server in order to avoid things going wrong and make sure plugins don't run. Or should block the server from loading until the user decides whether it should continue (Just like FML has the /fml confirm)

You're assuming that Sponge loads in some capacity, but I think much more likely scenarios for failure are that sponge simply isn't present or doesn't get loaded at all somehow. It's very likely that if Sponge loads *partially* then the game will just crash anyway, which isn't a problem. I don't think there's any need to over-complicate the startup process by trying to define what "loaded successfully" actually means and trying to somehow measure it.

src/main/java/org/spongepowered/mod/mixin/entity/projectile/MixinEntityArrow.java
```diff
@@ -111,4 +114,16 @@ public void setCritical(boolean critical) {
     public MixinEntityArrow(World worldIn) {
         super(worldIn);
     }
+
+    @Override
+    public void readFromNbt(NBTTagCompound compound) {
```
gabizou: I'm fairly sure we can use at annotation to mix in to the target method

simon816: @gabizou It does, [see here](https://github.com/SpongePowered/Sponge/pull/173/files#diff-d1e54a67a6e6054ffea2c372f361a87cR394)

ST-DDT:
Maybe we should state in the javadocs that custom data arent shared between different Sponge Server impls.

src/main/java/org/spongepowered/mod/interfaces/IMixinEntity.java
```diff
@@ -35,4 +36,9 @@
     Entity getTeleportVehicle();
 
     void setTeleportVehicle(Entity entity);
+
+    void readFromNbt(NBTTagCompound compound);
```
bloodmc: This is not needed. Use the supermixins branch in Mixin then use SoftOverride annotation instead.

bloodmc:
Please refactor this PR using the feature/supermixin branch in Mixin.

simon816:
I've updated it to use the supermixin functionality.
There's a few javadoc comments in `MixinEntity` to explain what the different methods are for.

simon816:
I've added the same methods to `MixinTileEntity` too, even though looking through the current tile entity mixins there's no issues, it's there for future use.

PR #174, opened 2015-02-15
Implement PlayerChatEvent

Aaron1011:


bloodmc:
This was already implemented

https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/mixin/event/player/MixinEventPlayerChat.java

PR #181, opened 2015-02-26
Feature/sdk

AbrarSyed:
The SDK is built by running ```gradle build``` or ```gradle sdk```. The output is found in ```build/distributions/``` unlike the majority of the other jar artifacts that are found in ```build/libs```

# This SDK exists to aid in setting up a workspace in which a sponge plugin can easily be tested!

Things that need to be decided and finalized
 - [ ] Add dependency on sponge  (needs debate and a decision)
 - [ ] Add a README (someone write this, or il copy and modify the forge one)
 - [ ] Add a proper example plugin (atm its just an example forge mod)
 - [x] add the gradle wrapper

This is a living pull-request, I WILL rebase and stuff before merging it into master.


progwml6:
the wrapper should probably be included in the SDK

Zidane:
There will never be a dependency on Sponge, the mod. The SDK should be solely for the API.

progwml6:
@Zidane in this case this should probably be move to the API repo

sdk/src/main/java/com/example/examplemod/ExampleMod.java
```diff
@@ -0,0 +1,20 @@
+package com.example.examplemod;
+
+import net.minecraft.init.Blocks;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.common.Mod.EventHandler;
+import net.minecraftforge.fml.common.event.FMLInitializationEvent;
+
+@Mod(modid = ExampleMod.MODID, version = ExampleMod.VERSION)
```
Zidane: @Plugin...

sdk/src/main/java/com/example/examplemod/ExampleMod.java
```diff
@@ -0,0 +1,20 @@
...
+public class ExampleMod
+{
+    public static final String MODID = "examplemod";
+    public static final String VERSION = "1.0";
+    
+    @EventHandler
```
Zidane: @SubscribeEvent
public void init(InitizationEvent event)

sdk/src/main/java/com/example/examplemod/ExampleMod.java
```diff
@@ -0,0 +1,20 @@
...
+
+import net.minecraft.init.Blocks;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.common.Mod.EventHandler;
+import net.minecraftforge.fml.common.event.FMLInitializationEvent;
+
+@Mod(modid = ExampleMod.MODID, version = ExampleMod.VERSION)
+public class ExampleMod
+{
+    public static final String MODID = "examplemod";
```
Zidane: PLUGINID

sdk/src/main/java/com/example/examplemod/ExampleMod.java
```diff
@@ -0,0 +1,20 @@
+package com.example.examplemod;
+
+import net.minecraft.init.Blocks;
+import net.minecraftforge.fml.common.Mod;
+import net.minecraftforge.fml.common.Mod.EventHandler;
+import net.minecraftforge.fml.common.event.FMLInitializationEvent;
+
+@Mod(modid = ExampleMod.MODID, version = ExampleMod.VERSION)
+public class ExampleMod
```
Zidane: ExamplePlugin

sdk/build.gradle
```diff
@@ -0,0 +1,50 @@
+buildscript {
```
Zidane: This gradle file shouldn't use FG at all if this is for a plugin for SpongeAPI...

AbrarSyed:
It seems some people are under the impression that this is an example build script for building a plugin using the API. No this is much more. This is so people can build against the API as well as test their plugin in a working MC environment.

Honestly you can build a plugin using the API with javac, we hardly need an SDK package for that...

Zidane:
@AbrarSyed 

Ah my mistake, in that case that is even better. Carry on :).

modwizcode:
How about eclipse and Intellij plugins? I was planning to work on an Intellij plugin to automatically create a sponge project.

AbrarSyed:
Intellij is actually easier than eclipse... Just run gradle idea in this SDK, and it will create a fully functioning project. For eclipse we have to ship a workspace.. Which we do in this case.

modwizcode:
I was talking about adding a "Sponge Project" type for intellij (and
eclipse) , so you get a full setup with each project. Which is something
the ide would need plugins for.

CrazyPyroEagle:
I think this should have multiple options... SpongeAPI (run with only SpongeVanilla and other plugins as dependencies) and Forge (run with Sponge, Forge, Forge mods and other plugins as dependencies).

PR #188, opened 2015-03-02
Updated mappings to snapshot_20150301

modwizcode:
This moves our mappings out of the 1.8 stone age so that Mixins aren't quite as cryptic.


modwizcode:
Implemented by https://github.com/SpongePowered/Sponge/commit/2e993e59d10ec2ff5906d9f1fb8cce79f585bb63

PR #191, opened 2015-03-04
Update for SpongeAPI changes

DDoS:
- Implements some of the new changes
- Fixes a copy-paste error in MixinChunk

See: https://github.com/SpongePowered/SpongeAPI/pull/484

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/world/MixinWorld.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/world/MixinWorld.java
@@ -48,6 +48,7 @@ import net.minecraft.world.storage.WorldInfo;
 import org.apache.commons.lang3.exception.ExceptionUtils;
 import org.apache.commons.lang3.reflect.ConstructorUtils;
 import org.spongepowered.api.block.BlockLoc;
+import org.spongepowered.api.block.BlockState;
 import org.spongepowered.api.effect.particle.ParticleEffect;
 import org.spongepowered.api.effect.sound.SoundType;
 import org.spongepowered.api.entity.Entity;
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/mod/mixin/core/world/MixinWorld.java
```diff
@@ -182,12 +184,39 @@ public String getName() {
...
+
+    @Override
+    public void setBlock(int x, int y, int z, BlockState block) {
+        net.minecraft.world.World world = ((net.minecraft.world.World) (Object) this);
+        // 0 is no notify flag. For now not going to notify nearby blocks of update.
+        if (block instanceof IBlockState) {
+            world.setBlockState(new BlockPos(x, y, z), (IBlockState) block, 0);
+        } else {
+            // TODO: Need to figure out what is sensible for other BlockState implementing classes.
+            throw new UnsupportedOperationException();
```
ST-DDT: There should be some kind of error message here.

DDoS: That was a copy paste from BlockWrapper#getState(). I can add it to both maybe.

me4502:
May also be worth hooking the new NeighborNotifyEvent from Forge to BlockUpdateEvent

DDoS:
Could you elaborate on this please? The goal of the PR is just to update for my changes in the API, it's not really for new implementation.

me4502:
That was just modified in the api to be easier to implement

DDoS:
Don't merge yet, I can reduce some code duplication.

DDoS:
Fixed!

me4502:
Merged.

PR #193, opened 2015-03-06
[WIP] Implement ProjectileSource. FIXES #184

### LINKED ISSUE #184: ProjectileSource implementations lacking
I haven't found anyone implement `ProjectileSource` into `EntityPlayer` and other sources. This needs some fixing in order to spawn projectiles correctly.

me4502:
Work in progress implementation of ProjectileSource.

### DIFF
--- a/src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
+++ b/src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
@@ -253,6 +253,7 @@ import org.spongepowered.mod.text.translation.SpongeTranslation;
 import org.spongepowered.mod.weather.SpongeWeather;
 import org.spongepowered.mod.world.SpongeDimensionType;
 
+import javax.swing.text.html.Option;
 import java.awt.Color;
 import java.awt.image.BufferedImage;
 import java.io.File;
@@ -548,6 +549,15 @@ public class SpongeGameRegistry implements GameRegistry {
             id = "minecraft:" + id;
         }
         return Optional.fromNullable((EntityType) this.entityIdToTypeMappings.get(id));
+    }
+
+    public Optional<EntityType> getEntity(Class<? extends org.spongepowered.api.entity.Entity> entityClass) {
+        for(EntityType type : getEntities()) {
+            if(entityClass.isAssignableFrom(type.getEntityClass())) {
+                return Optional.of(type);
+            }
+        }
+        return Optional.absent();
     }
 
     @Override
--- a/src/main/java/org/spongepowered/mod/util/SpongeHooks.java
+++ b/src/main/java/org/spongepowered/mod/util/SpongeHooks.java
@@ -457,6 +457,7 @@ public class SpongeHooks {
             if(velocity != null) {
                 entity.setVelocity(velocity.toDouble());
             }
+            world.spawnEntityInWorld((net.minecraft.entity.Entity) entity);
             return entity;
         } catch(Exception e) {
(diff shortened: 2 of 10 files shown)


src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -279,6 +280,15 @@ public TradeOfferBuilder getTradeOfferBuilder() {
         return Optional.fromNullable((EntityType) this.entityIdToTypeMappings.get(id));
     }
 
+    public Optional<EntityType> getEntity(Class<? extends org.spongepowered.api.entity.Entity> entityClass) {
+        for(EntityType type : getEntities()) {
```
Aaron1011: Can this just be done by calling `entityClass.getType`?

The method `getType` is inherited from `EntityState`

Deamon5550: >Class.getType()

Just read that one to yourself quickly and see your issue :)

simon816: Why not just use the `entityClassToTypeMappings` field?

Aaron1011: @Deamon5550: See https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/entity/EntityState.java#L40

Soren025: @Aaron1011 see http://docs.oracle.com/javase/6/docs/api/

Soren025: See http://docs.oracle.com/javase/7/docs/api/java/lang/Class.html

me4502: @simon816 In my testing that didn't work correctly. Probably because the class object is talking about the MC object that the interface was mixed in to, and as we are just passing the interface, it doesn't notice it.

src/main/java/org/spongepowered/mod/util/SpongeHooks.java
```diff
@@ -417,4 +427,21 @@ public static SpongeConfig getActiveConfig(World world) {
...
+            entity.setShooter(source);
+            if(velocity != null) {
+                entity.setVelocity(velocity.toDouble());
+            }
+            return entity;
+        } catch(Exception e) {
+            SpongeMod.instance.getLogger().error(ExceptionUtils.getStackTrace(e));
+        }
+
+        return null;
```
simon816: Returning `null` seems like a bad idea.

simon816:
Apart from `MixinEntityBlaze` and `MixinEntityGhast`, the mixed-in methods are identical.
To improve maintainability as well as reduce code duplication, you could target multiple entity classes. Which is what I did in [`MixinArmorEquipable`](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/mixin/core/entity/MixinArmorEquipable.java)

me4502:
@simon816 
Are you able to assist with this PR? For example, making it work?

simon816:
@me4502 See this commit for my idea https://github.com/simon816/Sponge/commit/b022cb4cd7842faf91ff8db00c8b92c21ae9e531

simon816:
@me4502  I have moved this to SpongeCommon on the branch [`feature/projectilesource`](https://github.com/SpongePowered/SpongeCommon/compare/feature/projectilesource).
Implementation can continue there.

PR #195, opened 2015-03-06
Implement EntityInteractionType methods

Aaron1011:


src/main/java/org/spongepowered/mod/entity/SpongeEntityInteractionType.java
```diff
@@ -0,0 +1,75 @@
...
+            return false;
+        }
+        SpongeEntityInteractionType other = (SpongeEntityInteractionType) obj;
+        if (this.name != other.name) {
+            return false;
+        }
+        return true;
+    }
+
+    public DataContainer toContainer() {
```
simon816: What are these serialization methods doing here?

Aaron1011: They're present in `SpongeEntityMeta`, which has the same purpose as this class (except it also requires a `type`, something unnecessary for this).

Those methods should be removed from `SpongeEntityMeta` as well if they don't belong here.

simon816: I don't see why these methods are here or in `SpongeEntityMeta`, if there's no `@Override` annotation, it should not be included.

me4502: I'd probably just say remove them from this PR for now.

src/main/java/org/spongepowered/mod/entity/SpongeEntityInteractionType.java
```diff
@@ -0,0 +1,75 @@
...
+        this.name = name;
+    }
+
+    @Override
+    public String getName() {
+        return this.name;
+    }
+
+    @Override
+    public boolean equals(Object obj) {
```
Minecrell: Why do you override `equals` here? In most cases our "fake enum types" should be also checkable with reference equality - that means there is no need to override this method.

src/main/java/org/spongepowered/mod/entity/SpongeEntityInteractionType.java
```diff
@@ -0,0 +1,75 @@
...
+        return null;
+    }
+
+    public void serialize(DataSource source) {
+        // TODO
+    }
+
+    @Override
+    public String toString() {
+        return MoreObjects.toStringHelper(this)
```
Minecrell: We don't have `MoreObjects` on 1.8, use `Objects.toStringHelper` instead

Aaron1011:
Closed in https://github.com/SpongePowered/Sponge/commit/a9d4cab22bf394d77b6f9aa175279afa65b52326

PR #197, opened 2015-03-09
Properly implement isInWater for boats

Aaron1011:
This PR implements `isInWater` for boats by switching to the Vanilla code used in `onUpdate` for determining if a boat should be considered in water.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/entity/vehicle/MixinEntityBoat.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/entity/vehicle/MixinEntityBoat.java
@@ -144,8 +144,8 @@ public abstract class MixinEntityBoat extends Entity implements Boat {
 
         double y1 = this.getEntityBoundingBox().minY + (this.getEntityBoundingBox().maxY - this.getEntityBoundingBox().minY) * (double)(maxLoop + 0) / (double)4.875;
         double z2 = this.getEntityBoundingBox().minY + (this.getEntityBoundingBox().maxY - this.getEntityBoundingBox().minY) * (double)(maxLoop + 1) / (double)4.875;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/mod/mixin/core/entity/vehicle/MixinEntityBoat.java
```diff
@@ -127,10 +129,27 @@ public MixinEntityBoat(World worldIn) {
...
     public boolean isInWater() {
-        // TODO This only works when the boat is submerged
-        return this.inWater;
+        byte var1 = 5;
+        int maxLoop = 4;
+
+        double var5 = this.getEntityBoundingBox().minY + (this.getEntityBoundingBox().maxY - this.getEntityBoundingBox().minY) * (double)(maxLoop + 0) / (double)var1 - 0.125D;
+        double var7 = this.getEntityBoundingBox().minY + (this.getEntityBoundingBox().maxY - this.getEntityBoundingBox().minY) * (double)(maxLoop + 1) / (double)var1 - 0.125D;
+        AxisAlignedBB var9 = new AxisAlignedBB(this.getEntityBoundingBox().minX, var5, this.getEntityBoundingBox().minZ, this.getEntityBoundingBox().maxX, var7, this.getEntityBoundingBox().maxZ);
+        if(this.worldObj.isAABBInMaterial(var9, Material.water)) {
```
ST-DDT: You should remove the if and return the result directly. That way it would be more readable, although it is not that close to the original code.
Why do you use variables without a proper name?

modwizcode:
Fix the variable names if you can, otherwise it looks good.

src/main/java/org/spongepowered/mod/mixin/core/entity/vehicle/MixinEntityBoat.java
Aaron1011: I was trying to keep it close to the way it appears in vanilla - I'll change it.

PR #199, opened 2015-03-12
Implement SQL service

zml2008:
Just want to PR this for some comments on the supported database platforms.
This differs from Bukkit in a few ways

1. Mariadb rather than MySQL connector: I chose the MariaDB one over the MySQL one because to me it seems like there has been a switch away from oracle MySQl based on the fact that Oracle sucks. (they should be protocol-compatible anyways)

2. H2 rather than sqlite: SQLite has a ton of limitations in a lot of areas (like modifying tables) that make working with it hard. H2 has a richer feature set, and as a bonus is pure-java. A few bukkit plugins have used this database engine in the past, so I think it is worth including instead of sqlite.

(and yeah, I dumped another slightly unrelated commit in because I had to work with the config while implementing this API)

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeMod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeMod.java
@@ -49,12 +49,14 @@ import org.spongepowered.api.plugin.PluginContainer;
 import org.spongepowered.api.service.ProviderExistsException;
 import org.spongepowered.api.service.command.CommandService;
 import org.spongepowered.api.service.command.SimpleCommandService;
+import org.spongepowered.api.service.sql.SqlService;
 import org.spongepowered.mod.command.CommandSponge;
 import org.spongepowered.mod.event.SpongeEventBus;
 import org.spongepowered.mod.event.SpongeEventHooks;
 import org.spongepowered.mod.guice.SpongeGuiceModule;
 import org.spongepowered.mod.plugin.SpongePluginContainer;
 import org.spongepowered.mod.registry.SpongeGameRegistry;
+import org.spongepowered.mod.service.sql.SqlServiceImpl;
 import org.spongepowered.mod.util.SpongeHooks;
 
 import java.io.File;
@@ -86,8 +88,13 @@ public class SpongeMod extends DummyModContainer implements PluginContainer {
             SimpleCommandService commandService = new SimpleCommandService(this.game.getPluginManager());
             this.game.getServiceManager().setProvider(this, CommandService.class, commandService);
             ((SpongeEventBus) this.game.getEventManager()).register(this, commandService);
-        } catch (ProviderExistsException e1) {
-            logger.warn("Non-Sponge CommandService already registered: " + e1.getLocalizedMessage());
+        } catch (ProviderExistsException e) {
+            logger.warn("Non-Sponge CommandService already registered: " + e.getLocalizedMessage());
+        }
+        try {
+            game.getServiceManager().setProvider(this, SqlService.class, new SqlServiceImpl());
+        } catch (ProviderExistsException e) {
+            logger.warn("Non-Sponge SqlService already registered: " + e.getLocalizedMessage());
         }
     }
 
--- a/build.gradle
+++ b/build.gradle
@@ -79,7 +79,10 @@ dependencies {
     compile 'org.slf4j:slf4j-api:1.7.7'
     compile 'org.apache.logging.log4j:log4j-slf4j-impl:2.0-beta9'
     compile 'com.google.inject:guice:4.0-beta5'
-    
+    compile 'com.zaxxer:HikariCP-java6:2.1.0'
+    compile 'org.mariadb.jdbc:mariadb-java-client:1.1.8'
+    compile 'com.h2database:h2:1.4.186'
+    compile 'org.xerial:sqlite-jdbc:3.8.+'
     compile project('Mixin')
     compile project('SpongeAPI')
     
@@ -105,6 +108,10 @@ shadowJar  {
         include(dependency('ninja.leaping.configurate:configurate-hocon'))
         include(dependency('ninja.leaping.configurate:configurate-core'))
         include(dependency('com.googlecode.concurrentlinkedhashmap:concurrentlinkedhashmap-lru'))
+        include(dependency('com.zaxxer:HikariCP-java6'))
+        include(dependency('org.mariadb.jdbc:mariadb-java-client'))
+        include(dependency('com.h2database:h2'))
+        include(dependency('org.xerial:sqlite-jdbc'))
         include(project('SpongeAPI'))
         include(project('Mixin'))
     }
--- a/src/main/java/org/spongepowered/mod/command/CommandSponge.java
+++ b/src/main/java/org/spongepowered/mod/command/CommandSponge.java
@@ -90,7 +90,7 @@ public class CommandSponge extends CommandBase {
     public void processCommand(ICommandSender sender, String[] args) {
         if (args.length > 0) {
             String command = args[0];
-            SpongeConfig config = CoreMixinPlugin.getGlobalConfig();
+            SpongeConfig<?> config = CoreMixinPlugin.getGlobalConfig();
 
             if (COMMANDS.contains(command)) {
                 if (FLAG_COMMANDS.contains(command)) {
@@ -198,7 +198,7 @@ public class CommandSponge extends CommandBase {
                             }
                             configName = "ALL";
                         } else if (name.equals("*") && config.getType() == SpongeConfig.Type.DIMENSION) {
-                            for (SpongeConfig dimensionConfig : SpongeGameRegistry.dimensionConfigs.values()) {
+                            for (SpongeConfig<?> dimensionConfig : SpongeGameRegistry.dimensionConfigs.values()) {
                                 dimensionConfig.save();
                             }
                             configName = "ALL";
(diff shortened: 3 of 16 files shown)


zml2008:
@SpongePowered/developers it'd be good to get some comment on the choice of database engines

gratimax:
It's debatable whether Sponge itself should be opinionated with regards to database engines... on the previous PR, people started requesting a MongoDB service as well. Ideally this would be solved by some further abstractions in persistence (some of which have already been merged in) instead of creating a SQL engine, a MongoDB engine, a Couchdb engine...

ST-DDT:
It might be an outdated opinion but Its more likely to have a db server with MySQl or postgres instead of MariaDb served with your mc server or web server.

modwizcode:
Please do not bundle SQL, this is something that we can leave up to
developers of plugins.

src/main/java/org/spongepowered/mod/SpongeMod.java
```diff
@@ -86,8 +88,13 @@ public SpongeMod() {
...
             ((SpongeEventBus) this.game.getEventManager()).register(this, commandService);
-        } catch (ProviderExistsException e1) {
-            logger.warn("Non-Sponge CommandService already registered: " + e1.getLocalizedMessage());
+        } catch (ProviderExistsException e) {
+            logger.warn("Non-Sponge CommandService already registered: " + e.getLocalizedMessage());
+        }
+        try {
+            game.getServiceManager().setProvider(this, SqlService.class, new SqlServiceImpl());
+        } catch (ProviderExistsException e) {
+            logger.warn("Non-Sponge SqlService already registered: " + e.getLocalizedMessage());
```
simon816: I don't think it's even possible for another provider to be registered at this point.

AbrarSyed:
Seems you guys are a bit mistaken about mariadb.. It's a drop in replacement for mysql, an existing mysql connector should work completely fine on mariadb as well, I don't see why you guys are making a distinction.

As for the inclusion of this, I feel that lots of plugins will want to have access to an SQL database. And it would be a waste if all of them were to have to shade in their JDBC. For NoSQL DBs, I figure those will be rare and thus people can shade that themselves.

Mumfrey:
Indeed, reduction in shading of common dependencies reduces likelihood of library conflicts as well.

modwizcode:
Make official database services as separate plugins, packaged but not
shaded into the main server, then use those in plugins. If server owners
don't want the extra functionality they won't have to keep them.

progwml6:
we could also provide small bridge plugins that have common DB connectors in them to avoid these conflicts

AbrarSyed:
Various official DB module's a good idea. They would also serve as examples for people to implement their own.

ZephireNZ:
I agree with @gratimax on using DataSource for databases as well. We have an abstract system for serialization set up, so why not use that rather than exposing the bare JDBC connector? Then it would be up to the developer to either use the serialization (which would be the recommended way), or to use bare JDBC and leave it to the developer to shade in any dependencies they need. If most are using the built in serialization, there should hopefully be few developers actually shading in DB libraries.

modwizcode:
Can someone clarify what this PR is actually for? Is it related to the Sponge config only or is it supposed to be more general?

ZephireNZ:
AFAIK, it's for providing a recommended way of accessing a JDBC database
that is configured in the server configs. Bukkit also had this, but most
people did not use it.

The refactoring of config stuff seems to be a side effect of the changes
they've made.

zml2008:
It's a connection pool setup for JDBC connections that plugins can use. The config stuff isn't really related, the layout just bugged me.

We already debated the merits of having jdbc access when the PR for the API for SqlService happened (the abstractions just don't provide the power that a relational db does -- all ORMs suck). That's not what this PR is about. This PR is more about which SQL implementations should be shaded/promoted.

modwizcode:
Then I stand by my statement of none, make them official,
prepackaged/downloaded plugins that are optional and removable. That
prevents any issues we might have with missing providers/broken providers.

It's also more instep with our goals to expand peoples options by not
bundling this right into the server. That was one of Bukkit's mistakes,
best not to repeat it.

AbrarSyed:
I agree with modwizcode on this matter.

zml2008:
like anybody will actually read enough to download the necessary plugins for SQL severs. Not bundling a few common JDBC connectors just means that every plugin dev using SQL in their plugin will get more users screwing up the installation.

modwizcode:
You missed what's going on here, when someone gets Sponge they get Sponge +
plugins for SQL. If they don't have the plugins then they are actively
breaking the install.

AbrarSyed:
Then... What databases are we promoting providing?

I'd say just mysql and H2, swlite has been as pain in my experience. Maybe mongo as well...

zml2008:
well @AbrarSyed mongo is not JDBC so that's not in the scope of this PR
but yeah, mysql+H2 is what I currently have hooked up in this PR

and @modwizcode so now people have 3 jars instead of one?

ST-DDT:
> so now people have 3 jars instead of one?

@zml2008 Yes, but in case they don't  need sql at all, their server uses at least 1 MB less perm gen per artifact you want to include. I know that does not sound much, but it adds up without any reasononable benefit. Here a short summary of the benefits:

- Without jdbcs bundled
  - Less Perm gen used
  - Reduce startup time
  - reduce dependencies
  - Smaller jar file size/faster upload
  - More freedom to choose a db connector
  - Guaranteed no issues with licences
  - Ability to choose jdbc version independently from the server's
    - does not block server releases, if the database adapters aren't uptodate
  - Avoids conflicts if MC decides to bundle a connector themselfes
  - (Teaches good practises)
    - Avoid including impls in apis
- With jdbc bundled
  - less files
  - no need to care to download a connector

sk89q:
So there seems to be some massive misconceptions here...

Java comes with JDBC already. JDBC is a core API. You cannot bundle "JDBCs." You do not choose JDBC versions. When you place a connector on classpath, it appears as accessible classes. This action does not enforce a connector or "reduce freedom." Does having Apache Commons on classpath prevent you from using Guava? No.

This PR effectively puts some common connectors or embedded databases on classpath. It does not mandate their use. It does not even affect Sponge at all. The PR also puts a popular connection pool library on classpath. Once again, it is optional and does not affect Sponge. 

There is one thing that does affect Sponge directly in this PR: there is a SQL service that provides a frontend to this connection pool library. In my opinion, this is unnecessary because the SQL service adds basically nothing over the library. However, this does not reduce freedom either; you are still free to do whatever.

Having extra JARs on classpath does not automatically bloat PermGen. Your standard JRE does not automatically load every class it can find in every JAR it touches -- that would be incredibly slow. It does not really affect startup time because ZIP/JAR files store an index. It's true a bigger file would have a higher tendency to fragment on a filesystem, but with fairly big cluster/block sizes these days, a file is not going to be that fragmented.

Something like a connector for MySQL/MariaDB should be bundled IMO because a lot of people will use it, and bundling this very common library will collectively save everyone from a huge headache. Something like H2 or a connection pool is more debatable, but it is worth a discussion.

Bukkit did bundle a connector for MySQL/MariaDB, as well as SQLite and possibly H2. Bukkit also had Ebean (an ORM), and provided Ebean access through Bukkit's API. Bukkit did not provide any connection pooling. Ebean was complete trash and is one of the worst ever ORMs I have had the displeasure of trying to use, so that's likely why it never caught on.

me4502:
I agree with sk. I do think we should include an SQLite connector, as they are used a fair bit for local storage.

MyPictures:
I do agree with bundling database connectors to Sponge, which one to implement... I'm not sure.
@modwizcode having some modular structure is always nice but for something basic like a connector it doesn't make much sense if you ask me.

modwizcode:
I think we should have H2 and MariaDB/MySQL support built in, I simply
remember Bukkit making it a hassle to have H2 work properly. IIRC you had
to modify the launch classpath to add H2 there or something like that.
Which is not what I want for Sponge. As for permgen, Java 8 doesn't even
have that, so it's not so much of a concern going forward.

ST-DDT:
> As for permgen, Java 8 doesn't even have that, so it's not so much of a concern going forward.

Well we do not use java8 to compile because some servers might not have it or am i wrong?

sk89q:
PermGen -- where classes are stored -- is still there on Java 8, but it's been merged into a shared memory pool. You still need a place to store loaded classes.

You don't have to compile code to Java 8 to use it on Java 8.

zml2008:
Thanks for helping to clear things up sk. 

H2 was not included in Bukkit -- a few plugins still used it because it has far fewer limitations than sqlite.

My main reasoning behind providing a service to handle connection pooling is the fact that a lot of servers use one database for multiple plugins' data. A service means that every plugin using a database can share a connection pool, reducing load on the database server (and hinting to database-using plugins that perhaps a connection pool might be the appropriate way to go). 

On our side, hiding the details of establishing a connection behind a service means that the specific connection pool and database connector implementations we use (like mysql vs mariadb) can remain implementation details that we can easily change in the future. 

Regarding sqlite, I sorta don't want to include it by default because sqlite imposes so many limitations on databases. I can still see why people with existing plugins that have userbases of sqlite users would want to be able to use those, so I would be OK also shading an sqlite connector as long as its use is discouraged for new work.

modwizcode:
SQLite is a fine database to use, it might not always be the correct choice
for a specific project, but it's a great local database.

On Sat, Mar 14, 2015 at 3:24 PM, zml <notifications@github.com> wrote:

> Thanks for helping to clear things up sk.
>
> H2 was not included in Bukkit -- a few plugins still used it because it
> has far fewer limitations than sqlite.
>
> My main reasoning behind providing a service to handle connection pooling
> is the fact that a lot of servers use one database for multiple plugins'
> data. A service means that every plugin using a database can share a
> connection pool, reducing load on the database server (and hinting to
> database-using plugins that perhaps a connection pool might be the
> appropriate way to go).
>
> On our side, hiding the details of establishing a connection behind a
> service means that the specific connection pool and database connector
> implementations we use (like mysql vs mariadb) can remain implementation
> details that we can easily change in the future.
>
> Regarding sqlite, I sorta don't want to include it by default because
> sqlite imposes so many limitations on databases. I can still see why people
> with existing plugins that have userbases of sqlite users would want to be
> able to use those, so I would be OK also shading an sqlite connector as
> long as its use is discouraged for new work.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/SpongePowered/Sponge/pull/199#issuecomment-80702971>.
>



-- 

Starbuck Johnson, Java Developer
www.modwiz.com | starbuck@modwiz.com

MattBDev:
I think SQLite has some uses for plugins but it is not always the best one. The benefit to SQLite is that the server owner, if inexperienced, can setup a plugin without being confused.

zml2008:
You all realize h2 does the same file-backed database with the same setup as sqlite, right? The only reason I see to use sqlite is if compatibility is required with non-java applications or existing data

modwizcode:
Or if you have experience with SQLite, it's more likely someone will know
SQLite than H2.

zml2008:
Ok, I've updated this PR so the Sqlite connector is shaded in. 

Really, h2 and sqlite are almost identical -- and in general the SQL databases are all fairly similar, so there's not a ton to learn.

PR #200, opened 2015-03-15
Implement Difficulties

AIOOB:


me4502:
This is looking nice

Minecrell:
`EntityInteractionType` is already implemented (and more complete) in #195

src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -220,6 +225,17 @@
             .put("ADVENTURE", new SpongeGameMode())
             .put("SPECTATOR", new SpongeGameMode())
             .build();
+    private static final ImmutableMap<String, Difficulty> difficultyMappings = new ImmutableMap.Builder<String, Difficulty>()
+            .put("PEACEFUL", Difficulty.class.cast(EnumDifficulty.PEACEFUL))
```
Minecrell: No need to use reflection for casting here, you can just cast normally or to Object first if the compiler complains

PR #204, opened 2015-03-20
Implement Persistence Services

gabizou:
With the Persistence API being merged, it's time for the various `DataSerializable` objects to actually successfully serialize their data to a `DataContainer`.

This has some changes to achieve this including:

- Provide a new service that requires registration before world loading starts
- Using SuperMixins for TileEntities
- Update to latest API for simpler usage of DataViews

### DIFF
--- a/src/main/java/org/spongepowered/mod/entity/SpongeEntityMeta.java
+++ b/src/main/java/org/spongepowered/mod/entity/SpongeEntityMeta.java
@@ -27,6 +27,8 @@ package org.spongepowered.mod.entity;
 import com.google.common.base.MoreObjects;
 import org.spongepowered.api.service.persistence.DataSource;
 import org.spongepowered.api.service.persistence.data.DataContainer;
+import org.spongepowered.api.service.persistence.data.DataQuery;
+import org.spongepowered.api.service.persistence.data.MemoryDataContainer;
 
 public class SpongeEntityMeta {
 
@@ -53,15 +55,17 @@ public class SpongeEntityMeta {
         SpongeEntityMeta other = (SpongeEntityMeta) obj;
         if (this.type != other.type) {
             return false;
-        } else if (this.name != other.name) {
+        } else if (this.name.equals(other.name)) {
             return false;
         }
         return true;
     }
 
     public DataContainer toContainer() {
-        // TODO
-        return null;
+        DataContainer container = new MemoryDataContainer();
+        container.set(new DataQuery("id"), this.type);
+        container.set(new DataQuery("name"), this.name);
+        return container;
     }
 
     @Override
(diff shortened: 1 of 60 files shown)


me4502:
Nice.

simon816:
To improve maintainability, I think it could internally call `writeToNBT` and then convert to a `DataContainer`, that way you won't need to add any new `DataQuery`s if a new tag is added in vannila. It would also allow getting data from mod tile entities too.

src/main/java/org/spongepowered/mod/mixin/core/block/data/MixinTileEntityBeacon.java
```diff
@@ -40,13 +42,11 @@
 @NonnullByDefault
 @Implements(@Interface(iface = Beacon.class, prefix = "beacon$"))
 @Mixin(net.minecraft.tileentity.TileEntityBeacon.class)
-public abstract class MixinTileEntityBeacon extends TileEntityLockable implements IUpdatePlayerListBox, IInventory {
+public abstract class MixinTileEntityBeacon extends MixinTileEntity {
```
AIOOB: Shouldn't this extend MixinTileEntityLockable?

gabizou:
> think it could internally call writeToNBT and then convert to a DataContainer

No. While it's decent to ensure that all data about a TE is being serialized, we shouldn't be serializing the data twice over. 

> It would also allow getting data from mod tile entities too.

There are changes in SpongePowered/SpongeAPI#475 that will help with this. This initial implementation is groundwork for setting up DataSerializables and DataSerializableBuilders. With this, it will be very apparent as to how the Persistence API is supposed to be used.

Also, remember that custom mod TileEntities don't extend other vanilla TileEntities (except maybe Lockable), so bringing up the point of being able to use a pre-existing vanilla specific `SpongeBeaconBuilder` isn't quite possible since there is no common `NBTTagCompound` that can be read from all `TileEntity` objects.

src/main/resources/sponge_at.cfg
```diff
@@ -45,6 +45,9 @@ public net.minecraft.village.MerchantRecipe field_180323_f # rewardsExp
 
 public net.minecraft.block.state.BlockState$StateImplementation
 
+public-f net.minecraft.tileentity.TileEntity field_145855_i # nameToClassMap
```
Minecrell: Why do you remove final from the field? You don't seem to be assigning new values to it.

src/main/java/org/spongepowered/mod/entity/SpongeEntityMeta.java
```diff
@@ -53,19 +55,17 @@ public boolean equals(Object obj) {
         SpongeEntityMeta other = (SpongeEntityMeta) obj;
         if (this.type != other.type) {
             return false;
-        } else if (this.name != other.name) {
+        } else if (this.name.equals(other.name)) {
```
simon816: Should be not equals, missing `!`

src/main/java/org/spongepowered/mod/mixin/core/block/data/MixinTileEntity.java
```diff
@@ -43,6 +46,9 @@
     @Shadow
     public abstract BlockPos getPos();
 
+    @Shadow
+    public abstract void markDirty();
```
simon816: Why add the shadowed method here and not in `MixinTileEntityBanner` where it's called?

PR #205, opened 2015-03-21
Implement Text API 2

Minecrell:
This implements the complete newly designed Text API from SpongePowered/SpongeAPI#425. (only the implementation of the Selector API is currently still missing)

It is implemented using a few mixins for the Minecraft classes to cache instance of the Text API classes for Sponge.

For efficiency, it is also using mixins for the Text classes in the API:

- **Mixins for the API? Sounds more like an implementation leak of a bad designed API.**  
It general that would be true, but in this case using mixins for the Text API classes **is not necessary**, but simply makes the whole implementation more efficient and easier maintainable.
- **Why are the mixins needed?**  
The general problem here is we need a way to cache the internal Minecraft (chat component) instances for our Text API instances. Theoretically we could also store them in a weak hash map using a lot of code around, but immutable instances are a perfect place to cache things (as they will never change and can be initialized lazily once) and storing it directly in the Text class not just makes it more efficient, but is also easier maintainable than having a weak hash map laying around somewhere.

While we need to be extremly careful with adding mixins for the API (it should be just exceptions), in some cases they simplify the implementation quite a lot and provide additional efficiency [which is also one of our goals for our implementation](https://docs.spongepowered.org/en/faq.html#summary).


Zidane:
@Minecrell 

Getting this fixed up for master is a priority.

Minecrell:
@Zidane I would have fixed it already but I'm waiting for someone to update master to the latest SpongeAPI because there are tons of problems with the merged items and persistence API changes. I think #204 does something for this

Minecrell:
@gabizou Where are all the vector changes coming from? I'm quite sure I didn't add them

kenzierocks:
Upstream. I requested a submodule update, @gabizou fixed all the build errors.

Minecrell:
@kenzierocks Eh, they're kinda out of scope for the PR

kenzierocks:
True. It wasn't my idea.

gabizou:
There were some changes from the SpongePowered/SpongeAPI#221 that were pushed to master, and I felt they should be merged in with Text API being closer to getting merged into Sponge Master than having to make yet another PR for API updates.

Minecrell:
@gabizou Apparently everyone else is against this implementation as-is, so I guess it will take a while until we have an implementation in master

gabizou:
@Minecrell still can't push any API based changes (like vectors) until Text 2 is merged in though.

Minecrell:
@gabizou That's why I wanted to wait with the merge in SpongeAPI until the implementation is also ready

gabizou:
@Minecrell I didn't merge in Text API 2 :stuck_out_tongue_winking_eye: you have @Zidane to thank.

PR #206, opened 2015-03-22
[WIP] Implement explosion events

gFreezer:
I implemented the explosion events because I need them for a private little server project.
They are close to being finished, however there are still some problems I couldn't find a solution for:

 - I couldn't figure out how to change the right client constructor with MixinExplosion, so this only works on the dedicated server
 - Plugins should be able to control which blocks to destroy by removing items from the list returned by `getBlocks()`. Changes to this list don't change the explosion outcome currently
 - I was unsure on what `Cause` to return in `getCause()`


src/main/java/org/spongepowered/mod/mixin/core/event/entity/MixinEventEntityExplosion.java
```diff
@@ -0,0 +1,111 @@
...
+public abstract class MixinEventEntityExplosion extends ExplosionEvent implements EntityExplosionEvent {
+
+    @Shadow
+    public abstract List<BlockPos> getAffectedBlocks();
+
+    private List<BlockLoc> affectedBlocks;
+
+    public MixinEventEntityExplosion(net.minecraft.world.World world, Explosion explosion) {
+        super(world, explosion);
+        this.affectedBlocks = new ArrayList<BlockLoc>();
```
simon816: Instructions in the constructor are ignored. You can instead add this assignment to the field declaration.

src/main/java/org/spongepowered/mod/mixin/core/world/MixinExplosion.java
```diff
@@ -0,0 +1,104 @@
...
+    @Shadow
+    private float explosionSize;
+
+    @Shadow
+    private boolean isFlaming;
+
+    private float explosionYield;
+
+    //TODO: Make this work on client
+    @SideOnly(Side.SERVER)
```
kenzierocks: Why doesn't this get injected on the client? This makes it break for SP worlds entirely.

src/main/java/org/spongepowered/mod/mixin/core/world/MixinExplosion.java
```diff
@@ -0,0 +1,104 @@
...
+        this.explosionYield = 1.0F / explosionSize;
+    }
+
+    @ModifyArg(
+            method = "doExplosionB(Z)V",
+            at = @At(
+                    value = "INVOKE",
+                    target =
+                    "Lnet/minecraft/block/Block;dropBlockAsItemWithChance(Lnet/minecraft/world/World;Lnet/minecraft/util/BlockPos;Lnet/minecraft/block/state/IBlockState;FI)V"))
+    private
```
kenzierocks: Extra newline?

src/main/java/org/spongepowered/mod/mixin/core/world/MixinExplosion.java
gFreezer: The client has two additional constructors, so I had to add the arguments to the method descriptor. Seems like I made a mistake there which made me think you can't select a specific constructor.
Just tried it again, it's working now, will push it soon

src/main/java/org/spongepowered/mod/mixin/core/world/MixinExplosion.java
gFreezer: Whoops, the Eclipse formatter did something unintended there...

src/main/java/org/spongepowered/mod/mixin/core/event/entity/MixinEventEntityExplosion.java
gFreezer: I thought that, but then I saw [MixinBlockUpdateEvent](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/mixin/core/event/block/MixinBlockUpdateEvent.java#L52).
I wanted to already fill the list here (see `getBlocks()`) then.

simon816: `MixinBlockUpdateEvent` is actually incorrect, none of the instructions are added to the target class. (Can be seen by looking at the output of the mixin transformer).

In this case, you can make an injection to the constructor, `@Inject(method = "<init>, at = @At("RETURN"))`
and add the blocks there

bloodmc:
@gFreezer I took care of this in commit https://github.com/SpongePowered/Sponge/commit/9ec1b64c1b4d2462b8255fe1cfef90357191c073

Thanks for the support!

PR #207, opened 2015-03-22
Register the human entity as a mod entity

simon816:
### Update
The human entity now exists in SpongeCommon as an `EntityCreature`. Therefore this PR is almost not needed.
I have re-purposed this PR to register the human as a mod entity and cleanup `MixinEntityRegistry`.

### Old description
This implements the `Human` entity.

I have done some testing, so far there are no issues when a human and I am in the world. However, it is difficult to properly test how well this works when there are many players on the server + many humans.

The concept is simple, to the server, a human is just another entity, an `EntityLivingBase`.
To the client, a human is another player who is AFK, an `EntityOtherPlayerMP`.

I am looking for feedback on the implementation so far, and whether there are any concerns with it.

Due to `HumanEntity` extending `EntityPlayer`, some things have to be overridden so that it is not seen as a player on the server.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/fml/MixinEntityRegistry.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/fml/MixinEntityRegistry.java
@@ -24,26 +24,19 @@
  */
 package org.spongepowered.mod.mixin.core.fml;
 
-import com.google.common.collect.BiMap;
-import com.google.common.collect.ListMultimap;
 import net.minecraft.entity.Entity;
-import net.minecraft.entity.EntityList;
-import net.minecraftforge.fml.common.FMLCommonHandler;
-import net.minecraftforge.fml.common.FMLLog;
 import net.minecraftforge.fml.common.Loader;
 import net.minecraftforge.fml.common.ModContainer;
 import net.minecraftforge.fml.common.registry.EntityRegistry;
-import net.minecraftforge.fml.common.registry.EntityRegistry.EntityRegistration;
-import org.apache.logging.log4j.Level;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.asm.mixin.Mixin;
-import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.asm.mixin.injection.At;
+import org.spongepowered.asm.mixin.injection.Inject;
+import org.spongepowered.asm.mixin.injection.callback.CallbackInfo;
 import org.spongepowered.common.entity.SpongeEntityRegistry;
 import org.spongepowered.common.entity.SpongeEntityType;
 import org.spongepowered.mod.SpongeMod;
 import org.spongepowered.mod.registry.SpongeModGameRegistry;
-
-import java.util.Map;
 
 @NonnullByDefault
 @Mixin(value = EntityRegistry.class, remap = false)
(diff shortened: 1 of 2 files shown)


ST-DDT:
@simon816 Thanks for pushing this forward.
In the attached issue you mentioned a problem with custom names/skins?
Is that problem fixed?

simon816:
Custom names are fixed, though vanilla sets a limit to 16 characters.
I have not pursued getting skins to work because there's no API for it. However, I could create a skin API and try to implement it. (The comment about skins was based off looking at the code in the client, I have not tried to get player skins on humans)

src/main/java/org/spongepowered/mod/mixin/core/world/MixinWorld.java
```diff
@@ -308,6 +315,39 @@ public boolean spawnEntity(Entity entity) {
...
+        return spawnEntityInWorld(((net.minecraft.entity.Entity) entity));
+    }
+
+    // Copy of original except doesn't force loading if entity is a HumanEntity
+    @Overwrite
+    public boolean spawnEntityInWorld(net.minecraft.entity.Entity entityIn) {
+        if (!this.isRemote && (entityIn == null || (entityIn instanceof net.minecraft.entity.item.EntityItem && this.restoringBlockSnapshots))) {
+            return false;
+        }
+        int cx = MathHelper.floor_double(entityIn.posX / 16.0D);
```
me4502: Usually >> 4 is used in this case.

Outfenneced:
Looks like skins are tied to UUIDs, so most likely an existing UUID for a player with the skin would have to be passed in for skins to work as far as non-modded clients go, but I may be wrong.

src/main/java/org/spongepowered/mod/mixin/core/world/MixinWorld.java
simon816: Was just copy + paste directly from decompiled MC source, I will change it though.

me4502: I may be wrong, leave it as is for now.

simon816:
Majority of implementation moved to https://github.com/SpongePowered/SpongeCommon/pull/19

src/main/java/org/spongepowered/mod/mixin/core/world/MixinWorld.java
```diff
@@ -58,10 +64,11 @@
 import org.spongepowered.mod.world.gen.SpongeGeneratorPopulator;
 
 import java.io.File;
+import java.util.List;
 
 
 @NonnullByDefault
-@Mixin(net.minecraft.world.World.class)
+@Mixin(value = net.minecraft.world.World.class, priority = 1111)
```
Zidane: Default priority is...1000 I believe. You can do 1001 here

simon816: I could. I just used the same value that's in [mixins.forge.core.json](https://github.com/SpongePowered/Sponge/blob/master/src/main/resources/mixins.forge.core.json#L6)

src/main/java/org/spongepowered/mod/mixin/core/world/MixinWorld.java
```diff
@@ -95,6 +111,38 @@ public void onConstructed(ISaveHandler saveHandlerIn, WorldInfo info, WorldProvi
...
+     * Only difference is that this includes forge's changes.
+     */
+    @Overwrite
+    public boolean spawnEntityInWorld(net.minecraft.entity.Entity entityIn) {
+        if (!this.isRemote && (entityIn == null || (entityIn instanceof net.minecraft.entity.item.EntityItem && this.restoringBlockSnapshots))) {
+            return false;
+        }
+        int cx = MathHelper.floor_double(entityIn.posX / 16.0D);
+        int cz = MathHelper.floor_double(entityIn.posZ / 16.0D);
+        boolean force = entityIn.forceSpawn || (entityIn instanceof EntityPlayer && !(entityIn instanceof HumanEntity));
```
Zidane: You could inject before here and capture locals so no Overwrite is needed.

src/main/java/org/spongepowered/mod/mixin/core/world/MixinWorld.java
Mumfrey: Bit-shifting doubles is usually a bad idea. You'd usually shift if it were a long or integer coord.

src/main/java/org/spongepowered/mod/mixin/core/world/MixinWorld.java
simon816: When this was made there was no such thing as capturing locals. I'll see if I can get that to work

simon816:
I have updated this PR now that the human entity is implemented in SpongeCommon

Zidane:
This needs to be done in SpongeCommon (just put it in the EntityList)

simon816:
https://github.com/SpongePowered/SpongeCommon/commit/c21cd010a0e7b1bbeb3bb75011c9b1950812c50c

PR #208, opened 2015-03-23
Map remaining sounds

simon816:
This maps the sounds added in https://github.com/SpongePowered/SpongeAPI/pull/464
Also switched small/big on slime walk sound around to be consistent.


me4502:
This is looking good.

PR #214, opened 2015-03-30
[WIP] Implement custom world generator API

rutgerkok:
This is the first implementation of the custom world generator API, as proposed by https://github.com/SpongePowered/SpongeAPI/pull/522. The implementation has been tested lightly, I have created a test plugin [here](https://github.com/rutgerkok/SpongeGeneratorTest).

The first custom world generator for Sponge

Most of the API is implemented, except two (not so minor) things:

* Sponge is not yet reading the configuration files to learn which plugin to use for world generation. The only way to test a world generator plugin currently is using the world.setWorldGenerator method.
* On a similar note, the server.createWorld methods aren't implemented either.

The test plugin gets around this by adding a `/generator` command, that calls world.setWorldGenerator.

I have also implemented the methods in BiomeType. Most of them aren't really related to custom world generation, but they are needed for a good implementation of this API. I also needed to implement some buffers.


src/main/java/org/spongepowered/mod/event/SpongeEventHooks.java
```diff
@@ -52,4 +57,23 @@ public void onEntityDeathEvent(LivingDeathEvent event) {
         SpongeHooks.logEntityDeath(event.entity);
     }
 
+    @SideOnly(Side.SERVER)
+    @SubscribeEvent
+    public void onChunkPrePopulate(PopulateChunkEvent.Pre event) {
```
Zidane: Might be best to have ours run first so other mods can come behind and do some tweaking based on our results in the PRE event.

src/main/java/org/spongepowered/mod/util/gen/SpongeMutableBiomeArea.java
```diff
@@ -0,0 +1,116 @@
...
+    private void checkOpen() {
+        Preconditions.checkState(!this.detached, "trying to use buffer after it's closed");
+    }
+
+    public SpongeMutableBiomeArea(int startX, int startZ, int sizeX, int sizeZ) {
+        super(new byte[sizeX * sizeZ], startX, startZ, sizeX, sizeZ);
+    }
+
+    @Override
+    public String toString() {
```
ST-DDT: Move this method to the bottom as it does not contain any functional logic.

bloodmc:
@rutgerkok looks good so far. Don't worry about world creation, i'll be pushing my impl to a branch soon which implements the feature/world-creation branch in API.

src/main/java/org/spongepowered/mod/util/gen/AbstractBiomeArea.java
```diff
@@ -0,0 +1,96 @@
...
+    @Override
+    public BiomeType getBiome(int x, int z) {
+        checkRange(x, z);
+        BiomeType biomeType = (BiomeType) this.biomeById[(x - this.startX) | (z - this.startZ) << 4];
+        return biomeType == null ? BiomeTypes.OCEAN : biomeType;
+    }
+
+    @Override
+    public Vector2i getMinBound() {
+        return new Vector2i(this.startX, this.startZ);
```
DDoS: You should probably cache these vectors to prevent unnecessary object creation, since these methods might be called often. This applies to other similar getters in your changes.

gratimax: The impl uses vectors in a lot of places. perhaps it would be a good idea to create a global cache of vectors and a util class to create vectors from, i.e. `SpongeVectors.newVec2i(int, int)` and so on

kenzierocks: Guava would make it easy to do with it's `Cache` class as well, so it would be easy to add.

rutgerkok: I will change the code to cache the vectors in all buffers. A global cache would be up to someone else to implement. :smile:

src/main/java/org/spongepowered/mod/util/gen/SpongeMutableBiomeArea.java
rutgerkok: Will do so.

src/main/java/org/spongepowered/mod/util/gen/AbstractBiomeArea.java
DDoS: Yeah, I agree, a global vector cache is out of the scope of this PR, but could be beneficial.

gratimax: I'll try making that into a separate PR

bloodmc:
@rutgerkok I pushed my world creation branch. You can find it here

https://github.com/SpongePowered/Sponge/tree/feature/world-creation

Still need to fix up generator support but world creation should work fine.

rutgerkok:
I have now pushed my changes to implement the improved API. The world creation part is still left out of this implementation.

I haven't done anything yet to make the two PRs compatible - the last few hours were just chasing bugs caused by the `buffer.getMaxBounds()` becoming inclusive, as well as updating my test plugin.

PR #215, opened 2015-03-30
Feature/permissions

zml2008:
This finishes off implementing the permissions api.

The included implementation is designed to be fairly restrictive -- it doesn't support any contexts besides the global context (and therefore doesn't use the contexts calculators or anything).

I've also stuck in changes to the command architecture to make commands registered using the MC commands API are visible in the Sponge command registry -- this passes all command execution through the CommandService.

I've started some work on wrappers for command senders. CommandSources/ICommandSenders not under our control are going to be an issue if anybody has big plugins/mods alongside Sponge. We'll have to provide wrappers since we can't just mix in to an interface (I wish...) and try to provide generic subinterfaces of CommandSource (like LocationCommandSource -- that can be used for both player & commandblock)?).


src/main/java/org/spongepowered/mod/SpongeCoremod.java
```diff
@@ -48,7 +48,7 @@ public SpongeCoremod() {
     @Override
     public String[] getASMTransformerClass() {
         return new String[] {
-                MixinBootstrap.TRANSFORMER_CLASS,
+            MixinBootstrap.TRANSFORMER_CLASS,
```
ZephireNZ: This probably doesn't need to be changed, or could be added to the OCD list.

src/main/java/org/spongepowered/mod/SpongeMod.java
```diff
@@ -166,14 +183,30 @@ public void onPreInit(FMLPreInitializationEvent e) {
...
+        try {
+            this.registry.init();
+            if (!this.game.getServiceManager().provide(PermissionService.class).isPresent()) {
+                try {
+                    this.game.getServiceManager().setProvider(this, PermissionService.class, new SpongePermissionService());
+                } catch (ProviderExistsException e1) {
+                }
+            }
+        } catch (Throwable t) {
+            logger.error("On init", t);
```
ZephireNZ: Is this debug? Otherwise, probably best to give more descriptive errors.

src/main/java/org/spongepowered/mod/effect/particle/SpongeParticleHelper.java
```diff
@@ -139,7 +139,8 @@
             float note = ((SpongeParticleEffect.Note) effect).getNote();
 
             if (note == 0f) {
-                return Lists.<Packet>newArrayList(new S2APacketParticles(internal, true, (float) px, (float) py, (float) pz, (float) ox, (float) oy, (float) oz, 0f, count, extra));
+                return Lists.<Packet>newArrayList(new S2APacketParticles(internal, true, px, py, pz, (float) ox, (float) oy, (float) oz, 0f,
+                        count, extra));
```
ZephireNZ: More unrelated formatting changes here

zml2008: yeah, I kinda went through and did everything but working on top of my previous changes. It's in a separate commit tho, so just look at the more relevant commits.

simon816: Could push the formatting/checkstyle fixes directly to master?

zml2008: not really -- they modify some code changed in the previous commits

src/main/java/org/spongepowered/mod/command/MinecraftCommandWrapper.java
```diff
@@ -0,0 +1,159 @@
...
+    private final ModContainer activeMod;
+    private final ICommand command;
+
+    public MinecraftCommandWrapper(ModContainer activeMod, ICommand command) {
+        this.activeMod = activeMod;
+        this.command = command;
+    }
+
+    @Override
+    public boolean call(CommandSource source, String arguments, List<String> parents) throws CommandException {
```
Mumfrey: `i`, `j` and `s2`? Maybe `userNameIndex`, `executeCount` and `currentArg`?

src/main/java/org/spongepowered/mod/command/MinecraftCommandWrapper.java
```diff
@@ -0,0 +1,159 @@
...
+    @Override
+    public String getUsage() {
+        return this.command.getCommandUsage(null);
+    }
+
+    @Override
+    public List<String> getSuggestions(CommandSource source, String arguments) throws CommandException {
+        final List<String> retList = new ArrayList<String>();
+        this.command.addTabCompletionOptions((ICommandSender) source, arguments.split(" "), null);
+        return retList;
```
Mumfrey: `retList` is not actually used so it will always be empty. Should it be passed into the call to `addTabCompletionOptions`?

src/main/java/org/spongepowered/mod/SpongeMod.java
```diff
@@ -166,14 +183,31 @@ public void onPreInit(FMLPreInitializationEvent e) {
...
+            this.registry.init();
+            if (!this.game.getServiceManager().provide(PermissionService.class).isPresent()) {
+                try {
+                    this.game.getServiceManager().setProvider(this, PermissionService.class, new SpongePermissionService());
+                } catch (ProviderExistsException e1) {
+                    // It's a fallback, ignore
+                }
+            }
+        } catch (Throwable t) {
+            logger.error("On init", t);
```
simon816: If there's an error in the initialization phase, it's probably best to stop the server from loading any further. Things will likely break otherwise

Zidane: I agree with @simon816

src/main/java/org/spongepowered/mod/registry/SpongeGameRegistry.java
```diff
@@ -1479,7 +1479,7 @@ private void setTextFactory() {
     }
 
     private void setSelectors() {
-        /*try {
+        /*try {aaa
```
simon816: ?
Accidental I guess

src/main/java/org/spongepowered/mod/SpongeMod.java
zml2008: Before the change loading continued on error, but gave a line indicating that an error occurred without printing a stacktrace or any useful info. I think the appropriate place to deal with this properly might be FML.

PR #217, opened 2015-04-01
Instance Unification impl

Deamon5550:
Implements [SpongeAPI#526](https://github.com/SpongePowered/SpongeAPI/pull/526)

Proposal for unified isFlowerPot detection
=============

Whilst it is well understood at this point that a well-specified and suitably future-proof API for Minecraft will need to be appropriately decoupled from Minecraft internals to withstand future changes to the underlying platform as effectively as possible, there are some aspects of the underlying platform which are intrinsically immutable and thus safe to propagate up to the API layer.

In many cases it is fully possible to achieve high levels of abstraction within the API, as demonstrated by the recent additions of the Inventory API and Item Data API. However there are still places where abstraction simply does not provide any measurable benefit to the end user, and adhering more faithfully to the internal architecture of the underlying Minecraft platform has no immediate or developer-facing drawbacks.

Where we can achieve some tangible improvement however is in homogenising developer-facing interfaces in order to provide a consistent and intuitive aspect-oriented development paradigm, reinforced by rigid adherance to integrated vertical software development disciplines.

### The Chaos at the Heart of Minecraft

With Minecraft, what you see is not always what you get. Crowd-sourced names for internal API structures can often fall somewhat short of making any sense whatsoever, but still provide valuable insight for newcomers into Minecraft internals, albeit tainted by occasionally apocryphal or sometimes downright nonsensical assignment of meaning to something that the assignor didn't understand or couldn't be bothered to decipher. Nevertheless, these little curios persist in the deobfuscated Minecraft codebase, and their legacy is one of some entertainment and wry humor when the truth of a particular field or method's origins come to light.

The typical ways of dealing with these conflicts between the *status quo* and common sense resolve into two camps:

* Argue about it pretty much forever
* Accept the nonsense and move on

An example of the former is the most hotly contested field of all time `World.isRemote`, sorry I meant `World.isClient` or is that `World.isR...` okay, okay I'll stop. However one of the most notable examples of the latter is the somewhat controversially named `isFlowerPot` method in Minecraft's `Block` class.

This method and its somewhat peculiar naming was the subject of some discussion until a Mojangsta was nailed down and forced to cough up the original name, which turned out to be something like `handlesOwnClone`, however the charm of the original (kind of quirky and entertaining) `isFlowerPot` was just too strong, and the method maintains its apocryphal nomenclature to this day.

### ClassLoading and You - a ~~song of ice and fire~~ soliloquy in five parts

Everyone is aware that Sponge is powered by ~~black magic~~ Mixins, however Class Loading in Java is a complex topic and Mixins are unable to adequately solve every problem we might throw at them. Fundamental to the successful application of Mixins is the core idea that a mixin is used to implement some aspect of the defined API.

`isFlowerPot` thus presents a problem. How can we adequately represent whether or not something is a flowerpot or not if we don't adequately encode this crucial information in the developer-facing portions of the API?

The answer is: define this information explicitly in each interface.

Taking a purely java-based approach to solving this problem alleviates several issues which would otherwise be caused. Whilst we could of course inject the required code at runtime, this rapidly becomes a very expensive operation when we factor in the sheer number of classes involved, and also the potential conflicts it creates with other non-flowerpot-aware code.

The beauty of this extremely [elegant](https://github.com/SpongePowered/SpongeAPI/commits/hotfix/instance-checks?page=16) and efficient implementation approach is that it provides a unified and 100 percent intuitive way of determining whether literally **any** object is or is not a flowerpot at runtime, without introducing complex transformer logic or polluting existing mixins with unrelated code. This allows mixins to focus on their task of implementing literally everything, whilst `isFlowerPot` logic can be handled in pure Java at a fundamental level within both the API and the Sponge implementation.

### DIFF
--- a/src/main/java/org/spongepowered/mod/entity/SpongeHorseColor.java
+++ b/src/main/java/org/spongepowered/mod/entity/SpongeHorseColor.java
@@ -32,4 +32,9 @@ public class SpongeHorseColor extends SpongeEntityMeta implements HorseColor {
         super(color, name);
     }
 
+    @Override
+    public boolean isFlowerPot() {
+        return false == true;
+    }
+
 }
--- a/src/main/java/org/spongepowered/mod/SpongeCoremod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeCoremod.java
@@ -52,6 +52,10 @@ public class SpongeCoremod implements IFMLLoadingPlugin {
         };
     }
 
+    public boolean isFlowerPot() {
+        return false;
+    }
+
     @Override
     public String getModContainerClass() {
         return "org.spongepowered.mod.SpongeMod";
--- a/src/main/java/org/spongepowered/mod/SpongeGame.java
+++ b/src/main/java/org/spongepowered/mod/SpongeGame.java
@@ -24,8 +24,17 @@
  */
 package org.spongepowered.mod;
 
-import com.google.common.base.Optional;
+import javax.annotation.Nullable;
+import javax.inject.Inject;
+
 import net.minecraftforge.fml.common.FMLCommonHandler;
+
+import org.objectweb.asm.AnnotationVisitor;
+import org.objectweb.asm.ClassWriter;
+import org.objectweb.asm.FieldVisitor;
+import org.objectweb.asm.Label;
+import org.objectweb.asm.MethodVisitor;
+import org.objectweb.asm.Opcodes;
 import org.spongepowered.api.Game;
 import org.spongepowered.api.GameRegistry;
 import org.spongepowered.api.MinecraftVersion;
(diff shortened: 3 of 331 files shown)


me4502:
There are numerous formatting errors throughout this PR.

Deamon5550:
srry is first pr and i try my best plz accept @me4502

src/main/java/org/spongepowered/mod/entity/SpongeHorseColor.java
```diff
@@ -32,4 +32,9 @@ public SpongeHorseColor(int color, String name) {
         super(color, name);
     }
 
+    @Override
+    public boolean isFlowerPot() {
+        return false == true;
```
simon816: This is the correct and most efficient logic check. Please be consistent and apply `false == true` to all other methods.

PR #218, opened 2015-04-01
[WIP] Implement world creation.

bloodmc:
In order to generate a proper level.dat per world, every world will now
use a seperate save handler unlike Forge/Vanilla which use only one.
This is required in order to have world settings saved per world.

In order to create a custom world, you will be required to use our new
WorldBuilder. This allows you to create and load a world immediately or
simply generate a world for later use.

Note: When creating a world, it will NOT be loaded. The world folder is
created along with initial data. If you want the world to be available,
you must load it by calling loadWorld.

After a world is created, Sponge will register these worlds,
if enabled, during server startup so they can be loaded automatically with
the default worlds. This should help plugins such as MultiVerse which had
to support this themselves.

UUID support has been added and is now stored with the world data.

All custom world data will be located in a new level dat file called
"level_sponge.dat". We chose to store all additional world data in a
seperate file to avoid loss of data if a user launched the server in a
non-forge or sponge environment.

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeMod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeMod.java
@@ -29,7 +29,10 @@ import com.google.common.eventbus.EventBus;
 import com.google.common.eventbus.Subscribe;
 import com.google.inject.Guice;
 import com.google.inject.Injector;
+import net.minecraft.nbt.CompressedStreamTools;
+import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.server.MinecraftServer;
+import net.minecraftforge.common.DimensionManager;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.fml.common.DummyModContainer;
 import net.minecraftforge.fml.common.FMLCommonHandler;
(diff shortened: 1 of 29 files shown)


src/main/java/org/spongepowered/mod/SpongeMod.java
```diff
@@ -246,4 +260,42 @@ public String getId() {
...
+                            NBTTagCompound nbt = CompressedStreamTools.readCompressed(new FileInputStream(levelData));
+                            if (nbt.hasKey(getModId())) {
+                                NBTTagCompound spongeData = nbt.getCompoundTag(getModId());
+                                if (spongeData.hasKey("dimensionId") && spongeData.getBoolean("enabled")) {
+                                    for (Map.Entry<Class<? extends Dimension>, DimensionType> mapEntry : getSpongeRegistry().dimensionClassMappings
+                                            .entrySet()) {
+                                        if (mapEntry.getKey().getCanonicalName().equalsIgnoreCase(spongeData.getString("dimensionType"))) {
+                                            if (!DimensionManager.isDimensionRegistered(spongeData.getInteger("dimensionId"))) {
+                                                DimensionManager.registerDimension(spongeData.getInteger("dimensionId"),
+                                                        ((SpongeDimensionType) mapEntry.getValue()).getDimensionTypeId());
```
simon816: Way too much nesting. If this happens normally I would invert the if statements and keep the indentation level low.
e.g. 
```java
if (directoryListing == null) {
    return;
}
for (File child : directoryListing) {
    if (!child.isDirectory()) {
        continue;
    }
    [...]
}
```

PR #222, opened 2015-04-02
Unify Forge and Vanilla implementation

Minecrell:
Our Forge implementation and our Vanilla implementation are both using MCP, which means they share the same mappings and provide a very similar Minecraft environment to work on. Because we're using mixins to implement our SpongeAPI interfaces directly on the Minecraft classes, we can copy most of the mixins simply from the current Forge implementation to our Vanilla implementation.

Copying everything from the one project to the other project results in a lot of work, and keeping the copied (and slightly modified) files up to date with the original files means even a lot more work, and certain changes (e.g. bug fixes) may get even forgotten which would result in bugs in the Vanilla implementation.

Being able to copy a lot of code from one implementation to another means we would be also able to share code directly between the projects using a shared subproject. Often only minor changes are needed in the Forge implementation for Forge integration and only a few other small changes in the Vanilla implementation.

Unifying both projects will get us more maintainable code and will save a lot of time required for copying everything over to Granite in the long term.

## Changes in this PR
- The project was separated into 3 subprojects:

  - **Common**: The shared subproject between both of the implementations. This will only contain code working in Vanilla and represents the base for both implementations. When implementation in here gets designed properly we're able to share a big amount of the implementations.
  - **Forge**: The Forge mod implementation (SpongeForge or sponge-forge)
  - **Vanilla**: The Vanilla implementation (previously Granite, now SpongeVanilla or sponge-vanilla)

- Currently there is no code shared between the implementations yet. The current sources were added as-is (with only a few smaller modifications for the unification)

- Executing `gradle build` will build _both implementations_, `gradle :Forge:build` or `gradle :Vanilla:build` will only build _one_ of the implementations.

### Known problems
- Currently a temporary ForgeGradle extension is used until we have native support for Vanilla (hopefully coming with 2.0)

- Currently the complete Minecraft source needs to be decompiled separately for each subproject, even if there is just a small access transformer change. This takes a bit longer than decompiling just one project.

- Importing the project in Eclipse setup using Eclipse's Gradle integration has currently one minor problem:
    - Importing the project after generating the Eclipse files using `gradle eclipse` works fine.

    - Importing the project as Gradle project using Eclipse's Gradle integration is currently not adding a dependency required for starting Vanilla, for some reason it is excluding it due to the location on disk of the dependency. When importing the project using Eclipse's Gradle integration, the start dependency needs to be added manually for the Vanilla project if you want to start the Vanilla server. This will probably change with ForgeGradle 2.0, the Forge subproject is not affected.


Voltasalt:
+1 :)

Mumfrey:
Just to add an opinion, but I'm not convinced that keeping the (effectively 3) projects in a single repository is the right way to go about this. Personally I think it would make more sense to promote the *current* `Sponge` repository into effectively the "Common" you've identified above and then have separate repositories for Sponge-Forge and Sponge-Vanilla. I feel this would be cleaner (Sponge project is a common dependency of both downstream implementations) and also provide a better roadmap for the future for if additional implementations are added.

The main reasoning for splitting it would be to take advantage of the same phased development relationship we have between `SpongeAPI` and `Sponge` where the git submodule can be left to "drift" behind the head until implementation "catches up".

For example, with the 3 projects in one repository, if a something is changed in Common which needs to propagate to Forge and Vanilla impls, those changes must all be made in lock-step. Whereas with Common (aka `Sponge`) being a git submodule this is not a requirement and the teams can update their own code at their own pace (with the Vanilla project not needing to wait for Forge to be updated or vice versa). This might not be an issue in the short term, but given this opens up the scope for other derived implementations it certainly presents a valid issue.

This would allows a lot of the extant code to stay exactly where it is, preserving the history of the files themselves which can be extremely useful metadata at times. Re-committing every single file in the repository like this effectively breaks the history chain of *every* file at this point in development. In essence you're 100% to *blame* for every file going forward. A minor point but still avoided by not contaminating the repository in the way you're proposing.

Compartmentalising the projects also makes sense from a CI perspective, since PRs and other ongoing changes can be made to common and tested *per se* without the concern that a Forge or Vanilla-specific detail will break the build, those changes can then be propagated downstream with little fuss.

Don't get me wrong, the split is a great idea. But my personal opinion (and disclaimer, this is only my opinion) is that this is the wrong way of going about it.

nightpool:
@Mumfrey how many more implementations are likely to use MCP though, seeing as that is what is being factored out?

Mumfrey:
@nightpool, liteloader and blazeloader for two, but that's not the point, mappings are largely not a significant issue when FG is in play because we have tools like srg2source etc. The key aspects are the ones I outlined above re: development effort and built-in benefits of splitting down the repo in the first place.

PR #224, opened 2015-04-02
[WIP] Implementation of Map API

modwizcode:
I'm not particularly skilled at implementation especially the more delicate and fine tuned tools available, additionally maps are fairly spread around which makes this extremely difficult for me to implement.

This is currently just the ground work and it will not build due to other missing API's, the implementation of MixinMapData is complete except for updating which I don't know how to fix/complete. Additionally it uses `@Overwrite` to perform most of the work because I don't know how to use `@Redirect` and `@Inject`.

If one of @SpongePowered/developers that is more skilled could walk me through this that would be fairly beneficial since I would like to focus more on implementation.


modwizcode:
This is long out of date and will be changed on SpongeCommon anyway.

PR #225, opened 2015-04-03
Fix synchronous task delay

### LINKED ISSUE #220: Delayed scheduled task not executing after the specified delay
When scheduling a task with `SynchronousScheduler#runTaskAfter(Object plugin, Runnable task, long delay)`, the task is executed immediately, rather than after the specified delay.

I found this while developing a warps plugin, and so when I printed out the amount of ticks since the server started using `Server#getRunningTimeTicks()`, both when the command was sent and when the Runnable was run, the same value was outputted, indicating that there was indeed no delay.

RobertHerhold:
Before this fix, a ScheduledTask's timestamp would by default be set to 0 rather than the current server tick counter. This would cause the task to be fired immediately, rather than after the appropriate delay. This PR fixes this issue, correctly firing ScheduledTasks after the correct delay. If pulled, this would resolve #220.

### DIFF
--- a/src/main/java/org/spongepowered/mod/service/scheduler/SyncScheduler.java
+++ b/src/main/java/org/spongepowered/mod/service/scheduler/SyncScheduler.java
@@ -265,6 +265,7 @@ public class SyncScheduler implements SynchronousScheduler {
         if (nonRepeatingTask == null) {
             SpongeMod.instance.getLogger().warn(SchedulerLogMessages.CANNOT_MAKE_TASK_WARNING);
         } else {
+            nonRepeatingTask.setTimestamp(this.counter);
             resultTask = this.schedulerHelper.utilityForAddingTask(this.taskMap, nonRepeatingTask);
         }
 
@@ -319,6 +320,7 @@ public class SyncScheduler implements SynchronousScheduler {
         if (nonRepeatingTask == null) {
             SpongeMod.instance.getLogger().warn(SchedulerLogMessages.CANNOT_MAKE_TASK_WARNING);
         } else {
+            nonRepeatingTask.setTimestamp(this.counter);
             resultTask = this.schedulerHelper.utilityForAddingTask(this.taskMap, nonRepeatingTask);
         }
 
@@ -389,6 +391,7 @@ public class SyncScheduler implements SynchronousScheduler {
         if (repeatingTask == null) {
             SpongeMod.instance.getLogger().warn(SchedulerLogMessages.CANNOT_MAKE_TASK_WARNING);
         } else {
+            repeatingTask.setTimestamp(this.counter);
             resultTask = this.schedulerHelper.utilityForAddingTask(this.taskMap, repeatingTask);
         }
 
@@ -459,6 +462,7 @@ public class SyncScheduler implements SynchronousScheduler {
         if (repeatingTask == null) {
             SpongeMod.instance.getLogger().warn(SchedulerLogMessages.CANNOT_MAKE_TASK_WARNING);
         } else {
+            repeatingTask.setTimestamp(this.counter);
             resultTask = this.schedulerHelper.utilityForAddingTask(this.taskMap, repeatingTask);
         }
 


RobertHerhold:
I don't believe the Travis build failing is my fault. See the below snippet of the failure log.

```
FAILURE: Build failed with an exception.
* What went wrong:
Could not resolve all dependencies for configuration ':userDevPackageDepConfig'.
> Could not resolve net.minecraftforge:forge:1.8-11.14.1.1341.
  Required by:
      org.spongepowered:Sponge:1.8-1341-1.1DEV-0
   > Could not HEAD 'http://files.minecraftforge.net/maven/net/minecraftforge/forge/1.8-11.14.1.1341/forge-1.8-11.14.1.1341-userdev.jar'. Received status code 500 from server: Internal Server Error
   > Could not GET 'https://libraries.minecraft.net/net/minecraftforge/forge/1.8-11.14.1.1341/forge-1.8-11.14.1.1341.pom'. Received status code 403 from server: Forbidden

```

RobertHerhold:
The fourth build was just re-run and it passes now.

Minecrell:
@RobertHerhold Yep, I restarted it ;)

RobertHerhold:
@Minecrell Haha thanks! Wasn't sure if someone did it manually or if Travis just did it

RobertHerhold:
What's the status on this PR?

bloodmc:
@sibomots is this ready to be merged?

sibomots:
If people are hot to merge it, go ahead to merge it.

But I don't think the changes actually are what we want.

Some facts:

The tick counter retrieved from the server is essentially an absolute counter. The counter variable in the Scheduler is not treated as an absolute counter, but a relative counter from when the next time the synchronous task should fire.

In other words, if a synchronous task needs to fire every N ticks, the task's initial counter setting is set to zero.  Then each time the SyncScheduler is woken up by a new Server Tick Event, the counter is incremented.  When the Synchronous Task counter (relative) reaches the desired interval N, then the task fires and the counter is reset to Zero again.

This PR seems to miss that point.

Go ahead and merge it and when the test case plugins fail, we'll know why.

In the future, PR's should reference _some_ sample code to show "before and after" effects.

Thx.

Mumfrey:
@sibomots based on what you're saying it looks like this is the right **place** for this fix but the wrong actual fix. Since it would seem that the value needs to be initialised to the current value of `SyncScheduler`'s internal `counter` variable based on [this code](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/service/scheduler/SyncScheduler.java#L202).

Would it be fair to say it's more suitable from an encapsulation point of view to have `SyncScheduler` assign the timestamp to the task?

Either way, it's easy to see that unless (and I guess it's possible) `SyncScheduler`'s `count` is tied to `Server::getRunningTimeTicks()` somehow, that this fix is definitively wrong.

RobertHerhold:
@Mumfrey Re-committed with setting the timestamp to `SyncScheduler`'s `counter`, and it works as intended.

 @sibomots Not sure what you meant in your original comment about `SyncScheduler`'s `counter` being set to 0 again... don't see that anywhere.

You might notice I committed with the same message (screwed up a rebase). If you want, I can close the PR and re-open it with the desired changes

Mumfrey:
> You might notice I committed with the same message (screwed up a rebase). If you want, I can close the PR and re-open it with the desired changes

Just squash the commits.

RobertHerhold:
I looked at that, but ran into some issues... fixed now :)

Mumfrey:
@RobertHerhold there are still two (or maybe just one) outstanding issues:

* Your formatting is broken on the lines you've added, you have added tabs
* You do not set the count for all instantiations of the task (that I can see), only two of them. Is this correct?

RobertHerhold:
I'll fix the formatting when I get home, and as for the second issue, I had a momentary memory lapse, thinking I only needed to set the timestamp for the delayed tasks, which is not the case.

RobertHerhold:
@Mumfrey Fixed now.

me4502:
Is this ready for merging, @bloodmc @sibomots ?

RobertHerhold:
Any progress here?

Zidane:
@RobertHerhold 

I just tested this with master and the initial delay kicked in and the delay works each time.

http://pastebin.com/bF3AFw52

Log http://pastebin.com/mK2aK93R

RobertHerhold:
That's interesting... I'll have to try it again...

Zidane:
I'll close this until further feedback/bugs arise.

PR #226, opened 2015-04-03
Implement networking API

simon816:
This implements the network channel API.
It piggy-backs off the network system provided by FML by using `FMLEventChannel`s

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeMod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeMod.java
@@ -171,6 +171,8 @@ public class SpongeMod extends DummyModContainer implements PluginContainer {
             // Add the SyncScheduler as a listener for ServerTickEvents
             FMLCommonHandler.instance().bus().register(this);
 
+            FMLCommonHandler.instance().bus().register(this.game.getChannelRegistrar());
+
             if (event.getSide().isServer()) {
                 SpongeHooks.enableThreadContentionMonitoring();
             }
--- a/src/main/java/org/spongepowered/mod/SpongeModGame.java
+++ b/src/main/java/org/spongepowered/mod/SpongeModGame.java
@@ -30,13 +30,14 @@ import net.minecraftforge.fml.common.FMLCommonHandler;
 import org.spongepowered.api.GameRegistry;
 import org.spongepowered.api.GameState;
 import org.spongepowered.api.Platform;
-import org.spongepowered.api.network.ChannelRegistrar;
 import org.spongepowered.api.plugin.PluginManager;
 import org.spongepowered.api.service.ServiceManager;
 import org.spongepowered.api.service.event.EventManager;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.api.world.TeleportHelper;
 import org.spongepowered.common.SpongeGame;
+import org.spongepowered.common.network.SpongeNetworkManager;
+import org.spongepowered.mod.network.SpongeModNetworkManager;
 
 import java.io.File;
 
@@ -49,6 +50,8 @@ public final class SpongeModGame extends SpongeGame {
             SpongeGame.API_VERSION, SpongeGame.IMPLEMENTATION_VERSION);
 
     private GameState state = GameState.CONSTRUCTION;
+
+    private final SpongeModNetworkManager networkManager = new SpongeModNetworkManager();
 
     @Inject
     public SpongeModGame(PluginManager pluginManager, EventManager eventManager, GameRegistry gameRegistry,
@@ -76,7 +79,7 @@ public final class SpongeModGame extends SpongeGame {
     }
 
     @Override
-    public ChannelRegistrar getChannelRegistrar() {
-        throw new UnsupportedOperationException(); // TODO
+    public SpongeNetworkManager getChannelRegistrar() {
+        return this.networkManager;
     }
 }
--- a/src/main/java/org/spongepowered/mod/interfaces/IMixinNetPlayHandler.java
+++ b/src/main/java/org/spongepowered/mod/interfaces/IMixinNetPlayHandler.java
@@ -1,0 +1,33 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.mod.interfaces;
+
+import java.util.Set;
+
+public interface IMixinNetPlayHandler {
+
+    Set<String> getRegisteredChannels();
+
+}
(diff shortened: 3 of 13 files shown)


Mumfrey:
There doesn't seem to be any implementation of the REGISTER contract in this PR or any connection to FML's mechanism for same.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -225,4 +234,41 @@ public void shutdown(Text kickMessage) {
...
+    public void registerChannel(Object plugin, ChannelListener listener, String channel) throws ChannelRegistrationException {
+        checkNotNull(plugin, "plugin");
+        checkNotNull(listener, "listener");
+        checkNotNull(channel, "channel");
+        checkArgument(SpongeMod.instance.getGame().getPluginManager().fromInstance(plugin).isPresent(),
+                "Provided plugin argument is not a plugin instance");
+        if (IMixinServer.vanillaChannels.contains(channel)) {
+            throw new ChannelRegistrationException("Channel \"" + channel + "\" cannot be registered");
+        }
+        if (this.channelListeners.containsKey(channel)) {
```
Mumfrey: This will prevent multiple plugins listening to the same channel.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -225,4 +234,41 @@ public void shutdown(Text kickMessage) {
...
+            throw new ChannelRegistrationException("Channel \"" + channel + "\" is already registered");
+        }
+        this.channelListeners.put(channel, listener);
+    }
+
+    @Override
+    public List<String> getRegisteredChannels() {
+        return ImmutableList.copyOf(this.channelListeners.keySet());
```
Mumfrey: This will only return plugin-registered channels and won't return channels registered by other consumers.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -225,4 +234,41 @@ public void shutdown(Text kickMessage) {
...
+        checkNotNull(channel, "channel");
+        checkArgument(SpongeMod.instance.getGame().getPluginManager().fromInstance(plugin).isPresent(),
+                "Provided plugin argument is not a plugin instance");
+        if (IMixinServer.vanillaChannels.contains(channel)) {
+            throw new ChannelRegistrationException("Channel \"" + channel + "\" cannot be registered");
+        }
+        if (this.channelListeners.containsKey(channel)) {
+            throw new ChannelRegistrationException("Channel \"" + channel + "\" is already registered");
+        }
+        this.channelListeners.put(channel, listener);
```
Mumfrey: No check for validity of channel name?

Mumfrey: Also no indication that current REGISTER is now out of date and needs to be sent again.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinMinecraftServer.java
simon816: `ChannelRegistrar.registerChannel`
> `@throws ChannelRegistrationException If the channel is already registered`

Mumfrey: Why? That makes no sense.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinNetHandlerPlayServer.java
```diff
@@ -82,4 +93,14 @@ public void sendCustomPayload(Object plugin, String channel, byte[] data) {
         sendPacket(new S3FPacketCustomPayload(channel, new PacketBuffer(Unpooled.wrappedBuffer(data))));
     }
 
+    @Inject(method = "processVanilla250Packet", at = @At(value = "INVOKE", target = "checkThreadAndEnqueue", shift = At.Shift.AFTER),
+            cancellable = true)
+    public void onProcessCustomPacket(C17PacketCustomPayload packetIn, CallbackInfo ci) {
+        if (IMixinServer.vanillaChannels.contains(packetIn.getChannelName())) {
+            return;
+        }
+        ci.cancel();
```
Mumfrey: Whoa there, so what you're saying here is that if a plugin registers an channel then **no other consumers** get to handle it? There is **no way** this is an acceptable behaviour.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinNetHandlerPlayServer.java
```diff
@@ -82,4 +93,14 @@ public void sendCustomPayload(Object plugin, String channel, byte[] data) {
         sendPacket(new S3FPacketCustomPayload(channel, new PacketBuffer(Unpooled.wrappedBuffer(data))));
     }
 
+    @Inject(method = "processVanilla250Packet", at = @At(value = "INVOKE", target = "checkThreadAndEnqueue", shift = At.Shift.AFTER),
```
Mumfrey: This fails build because `checkThreadAndEnqueue` needs to be fully-qualified.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinMinecraftServer.java
simon816: Need to make an issue on the API then. Originally added in https://github.com/SpongePowered/SpongeAPI/pull/363

src/main/java/org/spongepowered/mod/mixin/core/server/MixinNetHandlerPlayServer.java
simon816: Only vanilla channels (`MC|*`) are beyond this point so I thought there's no point letting it continue when it's already checked that the received channel is not vanilla

simon816:
Guess I'll have to see what this `REGISTER` business is all about, didn't see any references to it in vanilla.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinNetHandlerPlayServer.java
Mumfrey: Other things use plugin channels!

Mumfrey:
It's in [the original spec](https://dinnerbone.com/blog/2012/01/13/minecraft-plugin-channels-messaging/), how did you think it was appropriate to implement something if you don't actually understand how it works in the first place?

gabizou:
You're also forgetting a lot of `checkNotNull`s throughout.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -225,4 +234,41 @@ public void shutdown(Text kickMessage) {
...
+    private final Map<String, ChannelListener> channelListeners = new HashMap<String, ChannelListener>();
+
+    @Override
+    public void registerChannel(Object plugin, ChannelListener listener, String channel) throws ChannelRegistrationException {
+        checkNotNull(plugin, "plugin");
+        checkNotNull(listener, "listener");
+        checkNotNull(channel, "channel");
+        checkArgument(SpongeMod.instance.getGame().getPluginManager().fromInstance(plugin).isPresent(),
+                "Provided plugin argument is not a plugin instance");
+        if (IMixinServer.vanillaChannels.contains(channel)) {
```
Mumfrey: Thinking about it, this doesn't make sense either, because it would prevent a plugin from listening to a vanilla channel.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinMinecraftServer.java
Mumfrey: Fixed: SpongePowered/SpongeAPI@b65088cc85ebe6f4a8aa2520476fc8d1ccdde6a8

simon816:
I believe I have fixed the above issues.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -225,4 +231,23 @@ public void shutdown(Text kickMessage) {
 
         initiateShutdown();
     }
+
+    @Override
+    public void registerChannel(Object plugin, ChannelListener listener, String channel) throws ChannelRegistrationException {
+        checkNotNull(plugin, "plugin");
+        checkNotNull(listener, "listener");
+        checkNotNull(channel, "channel");
+        if (channel.length() > 16) {
```
kashike: 20

Mumfrey: Nope, 16 is correct.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinMinecraftServer.java
```diff
@@ -225,4 +231,23 @@ public void shutdown(Text kickMessage) {
 
         initiateShutdown();
     }
+
+    @Override
+    public void registerChannel(Object plugin, ChannelListener listener, String channel) throws ChannelRegistrationException {
+        checkNotNull(plugin, "plugin");
+        checkNotNull(listener, "listener");
+        checkNotNull(channel, "channel");
+        if (channel.length() > 20) {
```
simon816: So is this 16 or 20? I say 20 if vanilla supports it correctly.

simon816:
Updated to move some parts to https://github.com/SpongePowered/SpongeCommon/pull/18

Zidane:
@simon816 

Couple of issues (after messing around in SinglePlayer)

* SpongeRawDataInboundHandler needs to be @ChannelHandler.Sharable
* NetHandlerPlayClient needs to implement PlayerConnection. This in itself is gonna kill this PR right here and now since PlayerConnection requires a Player object to be returned. AbstractClientPlayer does not implement Sponge's Player interface. Due let me know what you come up with to solve this.

simon816:
This PR is finished. If anyone has any comments please say as this will be merged soon

PR #232, opened 2015-04-09
[WIP] Unify Forge and Vanilla implementation

Minecrell:
**This is a continuation of #222 - based on @Mumfrey's feedback it was splitted into 3 repositories now instead of a single one with subprojects.** (Feel free to skip the next part if you've already read this on the previous PR)

Our Forge implementation and our Vanilla implementation are both using MCP, which means they share the same mappings and provide a very similar Minecraft environment to work on. Because we're using mixins to implement our SpongeAPI interfaces directly on the Minecraft classes, we can copy most of the mixins simply from the current Forge implementation to our Vanilla implementation.

Copying everything from the one project to the other project results in a lot of work, and keeping the copied (and slightly modified) files up to date with the original files means even a lot more work, and certain changes (e.g. bug fixes) may get even forgotten which would result in bugs in the Vanilla implementation.

Being able to copy a lot of code from one implementation to another means we would be also able to share code directly between the projects using a shared subproject. Often only minor changes are needed in the Forge implementation for Forge integration and only a few other small changes in the Vanilla implementation.

Unifying both projects will get us more maintainable code and will save a lot of time required for copying everything over to Granite in the long term.

#### Project structure

The project consists out of 4 parts now: [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/tree/feature/unify), Sponge (the Common module holding the shared code of the implementations), [the Forge implementation](https://github.com/Minecrell/SpongeForge) and [Granite](https://github.com/SpongePowered/Granite/tree/feature/unify) (the Vanilla implementation).
To provide an unified Gradle setup in all these projects, the Gradle project setup was split into reusable files in SpongePowered/SpongeAPI#557 which is used for SpongeAPI itself, as well as Sponge and our two implementations.

#### TODO
- [X] Setup project structure
  - [ ] Merging SpongePowered/SpongeAPI#557 requires making the implementation compile for all the API changes first, especially the changes for the Data API and game registry.
- [X] Move most independent code into the Sponge module (Thanks to @Voltasalt for help with this)
- [ ] Integrate/update with changes from [`4af541517c03`...`master`](https://github.com/SpongePowered/Sponge/compare/4af541517c03...master)
  - [ ] Add world creation changes
- [x] Figure out the best way to share as much (Mixin) code as possible in the Sponge module and implementations (see below)
- [x] Create new repository for Forge implementation (Currently at [Minecrell/SpongeForge](https://github.com/Minecrell/SpongeForge))
- [x] \(Update README)

#### Sharing code between the implementations
While there is an huge part of our implementation that does not depend (nor reference) Forge/FML _at all_, some parts reference FML/Forge for smaller things like getting a list of worlds or saving custom NBT data for entities. Splitting this code out completely into the Forge module would fix the issue _itself_  but would again lead to a lot of **code duplication**, with just a single/few method calls as difference.
Because of this we need to find ways to keep as much code in Common as possible while still keeping direct Forge integration for the Forge implementation.

- For methods in Forge classes I've added an [`SpongeImpl`](https://github.com/SpongePowered/Sponge/blob/common/src/main/java/org/spongepowered/common/SpongeImpl.java) interface implemented by the implementation modules which provides unified access while still maintaining Forge integration. Of course this interface would grow more and more together with our implementation..

- For methods/fields added by Forge in Minecraft classes ([example: this is currently breaking Granite at runtime](https://github.com/SpongePowered/Sponge/blob/common/src/main/java/org/spongepowered/common/mixin/core/entity/MixinEntity.java#L134-L135)) maybe we could add the used fields and methods added by Forge in the Vanilla implementation (using mixins) before the Common mixins are applied. (Not completely sure if that would even work, @Mumfrey will probably know if it would)

- We'll probably need to split the methods/fields of some mixins so implementation-specific part of the mixins get separated out into extra mixins in the implementation module itself. This could of course lead to some confusion when you just want to find the implementation of a specific method in an implementation because you would need to look in multiple locations.

- ... @Mumfrey As the author of the Mixin system you know it best, so maybe you have some good ideas how we can set this up. =)

### (Still) known problems
- Currently a temporary ForgeGradle extension is used until we have native support for Vanilla (hopefully coming with 2.0)


halvors:
@Minecrell Isn't it better to rename the Sponge repository to SpongeForge and then create a new one called Sponge? That way we keep all the history and branches.

progwml6:
@halvors that is not an option in consideration

progwml6:
@halvors that is not an option in consideration

halvors:
@progwml6 Why not?

Minecrell:
@halvors The new PR keeps Sponge as the Forge mod and SpongeVanilla/SpongeCommon for the other modules. Sponge will stay the Forge implementation

halvors:
@Minecrell That's nice. But what i was trying to say was just that you should rename Sponge to SpongeForge and create a new repo called Sponge for the common stuff :)

Minecrell:
@halvors We decided to keep the current name (Sponge) for the Forge implementation, @progwml6 can probably tell you why.

PR #234, opened 2015-04-11
Implement support for server-side localized text

zml2008:
Companion to SpongePowered/SpongeAPI#563

### DIFF
--- a/src/main/java/org/spongepowered/mod/text/SpongeTextFactory.java
+++ b/src/main/java/org/spongepowered/mod/text/SpongeTextFactory.java
@@ -39,6 +39,7 @@ import org.spongepowered.api.text.format.TextStyles;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.mod.text.format.SpongeTextColor;
 
+import java.util.Locale;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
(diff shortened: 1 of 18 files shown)


src/main/java/org/spongepowered/mod/text/SpongeTextFactory.java
```diff
@@ -54,6 +55,10 @@ public Text parseJson(String json) throws IllegalArgumentException {
         }
     }
 
+    public static Locale getDefaultLocale() { // TODO: Get this from the MC client?
+        return Locale.getDefault();
```
ST-DDT: It would be cool to have a config for this.
Especially for servers which are hosted on big server providers without the ability to change the systems locale.

zml2008: yeah, the server's default locale is in a separate option so that something like that can be done without a ton of difficulty.

ST-DDT:
@zml2008 Could you please show me where the server side translations (ex from resource bundles) are converted into something that is send to the client?
Something like this:
````java
Text.Translatable translatable = ...;
Translation translation = translatable.getTranslation();
if (translation instanceof ClientTranslation) {
    // ClientTranslation is a translation that will take place client side
    client.sendChatComponent(new TranslationTextComponent(
        translation.getId(), 
        wrapArguments(translatable.getArguments())
        ));
} else {
    // otherwise resolve the translation on the Server side
    client.sendChatComponent(new TranslationTextComponent(
    // client.sendChatComponent(new ArgumentsTextComponent(
        translation.get(client.mpplayer$getLocale()), 
        wrapArguments(translatable.getArguments())
        ));
}
````

zml2008:
here's where the translation happens: https://github.com/SpongePowered/Sponge/pull/234/files#diff-9c3cc48564d143b45fc6a560c9f753c9R47

and here's where the player requests the translation: https://github.com/SpongePowered/Sponge/pull/234/files#diff-d6fac8c79b179f2e91fbd409d4a8b28fR121

ST-DDT:
Thanks

PR #244, opened 2015-04-18
Use Gradle daemon, for faster builds; see https://gradle.org/docs/current/userguide/gradle_daemon.html

vorburger:


kenzierocks:
Not everyone wants a background process running on their computer that's as intensive as gradle. This is easy to accomplish in your own terminal if you always want the daemon.

Mumfrey:
Doesn't the setting just cause it to use the deamon *if it's running* though? I mean, if it force starts the deamon then that's not good but if it just means it can use the deamon if present then I don't see the harm.

Although surely people who want to use the deamon will just put it in their personal `gradle.properties` in that case anyway?

Minecrell:
I'm not sure how this is making builds faster in total. In development you're typically using the run configuration support of your IDE which doesn't use Gradle in background. And I doubt we're building Sponge often enough to gain performance through this instead of taking constant CPU in background.

kenzierocks:
This is an always on flag, only bypassed by `--no-daemon`.

Mumfrey:
Let's assume this is a no for now then, since anyone who wants to use the deamon can just enable the option in their own `gradle.properties`.

Evolution0:
Hey,
As I understood from your last message you were looking for some information , I think this is it <http://shardedalti.remodelinghotline.com/aeefs>
forsberganthony@yahoo.com

Evolution0:
Hi,
Please read this info, it means so much to me, here is the link <http://wow.sablaturaenterprises.com/uerauenx>

Very truly yours, forsberganthony@yahoo.com

PR #247, opened 2015-04-20
Implement ChunkLayout

DDoS:
See https://github.com/SpongePowered/SpongeAPI/pull/581

### DIFF
--- a/src/main/java/org/spongepowered/mod/world/storage/SpongeChunkLayout.java
+++ b/src/main/java/org/spongepowered/mod/world/storage/SpongeChunkLayout.java
@@ -1,0 +1,139 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/mod/world/storage/SpongeChunkLayout.java
```diff
@@ -0,0 +1,123 @@
...
+
+public final class SpongeChunkLayout implements ChunkLayout {
+
+    private static final Vector3i CHUNK_SIZE = new Vector3i(16, 256, 16);
+    private static final Vector3i CHUNK_MAX = new Vector3i(30000000, 256, 30000000).sub(1, 1, 1).div(CHUNK_SIZE);
+    private static final Vector3i CHUNK_MIN = CHUNK_MAX.negate().sub(1, 0, 1);
+    public static final SpongeChunkLayout instance;
+
+    static {
+        instance = new SpongeChunkLayout();
```
simon816: This could be in the field declaration rather than in a separate `static` block

src/main/java/org/spongepowered/mod/world/storage/SpongeChunkLayout.java
```diff
@@ -0,0 +1,123 @@
...
+
+    @Override
+    public Optional<Vector3i> moveToChunk(int x, int y, int z, Direction direction, int steps) {
+        return moveToChunk(new Vector3i(x, y, z), direction, steps);
+    }
+
+    @Override
+    public Optional<Vector3i> moveToChunk(Vector3i chunkCoords, Direction direction, int steps) {
+        if (direction.isSecondaryOrdinal()) {
+            throw new IllegalArgumentException("Secondary cardinal directions can't be used here");
```
simon816: _could_ use `checkArgument` here

src/main/java/org/spongepowered/mod/world/storage/SpongeChunkLayout.java
DDoS: You can do that? Didn't know.

Mumfrey: It actually compiles to exactly the same bytecode as well.

PR #256, opened 2015-04-30
Display a help message when running the jar directly and add support for headless self-install

Mumfrey:
This PR adds functionality to have the Sponge mod jar generate a help message if the user tries to run it directly, and also provides a self-install mechanism for headless servers designed to make the transition from other *"just run the jar"* platforms more straightforward.

### Generated Help Message

Running the sponge jar in the console now produces the following message:

    --------------------------------------------------------------------------------
    
      Oops! You attempted to run the Sponge Forge Mod directly!
    
      This jar file is a mod for Minecraft Forge. You can run this file as a Forge
      mod by installing Minecraft Forge and simply dropping the jar file into the
      Forge "mods" directory. You can get Minecraft Forge from:
    
        http://files.minecraftforge.net/
    
      Once you have downloaded and installed Forge, you can load Sponge as a mod.
    
    
      Alternatively, for headless server installations you can use this Sponge jar
      as a launcher for Minecraft Forge. To use the jar as a launcher, you can:
    
        1) Install Minecraft Forge and the dependencies by running:
    
            java -jar sponge-1.8-1371-2.1DEV-0.jar install
    
        2) Run Minecraft Forge with Sponge by running
    
            java -jar sponge-1.8-1371-2.1DEV-0.jar run {args}
    
    
      If you need more help, you should check out our forums:
    
        http://forums.spongepowered.org/
    
    --------------------------------------------------------------------------------

This is intended to really drive home the fact that this jar is a **Minecraft Forge Mod** and not a stand-alone server. It also tells users where to go to download Forge, and where to go to get help.

If the environment is detected as being graphical, then the following dialog is displayed instead:

The clickable link takes the user directly to the Minecraft Forge downloads page.

### Headless Self-Installation

For users used to bukkit, having to download and install Forge seperately is inconvenient, especially if the server owner only wishes to run Sponge. Whilst it would be perfectly possible to have the jar automatically self-install Forge if run directly, I feel it's important that the user *knows* they are effectively using the jar as a *launcher* for Forge, and thus encourage them to run an **install** step followed by a **run** step. I think it would also be useful at some point to add a **setup** step which grabs the file and also writes out a bootstrap script which can then be customised and copies the sponge mod jar into the *mods* folder, effectively putting the user back in the state they would have been if they'd installed Forge separately.

The self-installation process uses Apache Ivy and is thus quite robust and can be easily maintained going forward.

The self-installation also creates the same `libraries` folder structure as the Minecraft launcher and the Minecraft Forge installer for consistency.

### Benefits

These features present the following benefits:

* Catches users who don't read the instructions and just try to run the jar, explains what the jar is and also points them to resources they will need.

* Doesn't obfuscate the fact that Forge is in the mix, in fact it rubs it right in the user's face so that they are aware that this is a Forge mod and they're using Forge.

* Self-Install feature uses Apache Ivy and thus updating will use the local Ivy cache where appropriate, this means that even if the user deletes the whole `libraries` folder it can simply be re-deployed from the Ivy cache without having to re-download everything. It also makes maintenance of dependencies and transitive dependencies a breeze because Ivy is awesome.

* Because Forge is downloaded by the implementation, it is guaranteed to be the correct version.

* If the user downloads a new Sponge and runs the same process, the new Forge will automatically be downloaded but existing dependencies will be used straight from the Ivy cache

### META-INF Additions

I have also updated the manifest with the target Forge version and Forge build. The intention is that these data can be used to provide more helpful error messages in other crash messages such as the mixin failure messages.

### One-command deployment for platforms

The self-installer also includes a command `go` which executes the same tasks as `install` followed by `run`. This means that for server platforms and other shared hosting solutions there is no "set up" step, the jar can simply be executed and trusted to set up and run Sponge in a single pass.


Mumfrey:
Pinging @Zidane and @bloodmc

src/main/resources/org/spongepowered/launch/ivysettings.xml
```diff
@@ -0,0 +1,12 @@
+<?xml version="1.0" encoding="ISO-8859-1"?>
+<ivysettings>
+    <settings defaultResolver="sponge" />
+    <resolvers>
+        <chain name="sponge">
+			<ibiblio name="m2" m2compatible="true" />
```
progwml6: our repo should be added here and central should NOT be in this list at all... all libraries that are not hosted by mojang itself that are needed are mirrored in our repository

Mumfrey: Funny, that's what I started with but it seems like the identifiers for some of the artefacts are wrong because it wouldn't pull scala off the forge repo for some reason.

progwml6: the forge repo keeps all the libraries compressed we have the raw uncompressed versions in our repo

nightpool:
loving this. such a cool idea.

ZephireNZ:
This is only if it is run from commandline, right? It won't catch users who try and run the jar in a GUI environment?

Mumfrey:
Now it displays this if you start the jar by double-clicking on it in windows or macos:

kenzierocks:
So pretty.

ST-DDT:
This is a cool feature, maybe we should move it to the API (jar), then we could just overwrite the help text file in the Sponge. That way we have consistent help/startup-error messages/windows/bevahiour along all sponge jars. (SpongeAPI, SpongeCommon, SpongeVanilla, SpongeForge....)

gabizou:
You know... it would be great if you could provide the same Panel for [SpongeAPI's Main](https://github.com/SpongePowered/SpongeAPI/blob/master/src/main/java/org/spongepowered/api/util/InformativeMain.java)...

PR #264, opened 2015-05-06
Standardize use of getEntity() in mixins

SonarBeserk:
Corresponding api pr: https://github.com/SpongePowered/SpongeAPI/pull/643

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayer.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayer.java
@@ -27,6 +27,7 @@ package org.spongepowered.mod.mixin.core.event.player;
 import net.minecraft.entity.EntityLivingBase;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraftforge.event.entity.living.LivingEvent;
+import org.spongepowered.api.entity.living.Living;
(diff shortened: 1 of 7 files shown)


SonarBeserk:
If anyone is able to specific while the build failed do please say. 3am is a poor time to debug.

src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayer.java
```diff
@@ -45,19 +46,7 @@ public MixinEventPlayer(EntityLivingBase entity) {
...
-    @Override
-    public Player getLiving() {
-        return (Player) this.entityPlayer;
-    }
-
-    @Override
-    public Player getHuman() {
-        return (Player) this.entityPlayer;
-    }
+    public Living getLiving() {return (Player) this.entityPlayer;}
```
gratimax: I think that you need to make this `public Player getLiving()`, since that is what is required by the interface PlayerEvent. The same follows for other methods.

zml2008: also n.ed to work on your formatting

simon816: Don't squish the methods onto one line like this.
IDoNotLiek

PR #267, opened 2015-05-07
Update common, move service provider initialization to common

zml2008:
Implementation of SpongePowered/SpongeCommon#25
Implementation of SpongePowered/SpongeAPI#650

This is not much, just bumping the command and adjusting service registration so it is all done in common.


Minecrell:
Merged in https://github.com/SpongePowered/Sponge/commit/439b6b5157c8e619577d9f1ee304a8afcc8dd36b

PR #276, opened 2015-05-19
Updated PlayerInteractBlockEvent implementation

thomas15v:
I needed this for my plugin. So yeah I tested everything :smile:.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.mod.mixin.core.event.player;
 
+import com.flowpowered.math.vector.Vector3d;
 import com.google.common.base.Optional;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.util.BlockPos;
@@ -43,10 +44,11 @@ import org.spongepowered.asm.mixin.Shadow;
 import org.spongepowered.common.Sponge;
 import org.spongepowered.common.registry.SpongeGameRegistry;
 import org.spongepowered.common.util.VecHelper;
+import org.spongepowered.mod.interfaces.IMixinEventPlayerInteractBlock;
 
 @NonnullByDefault
 @Mixin(value = net.minecraftforge.event.entity.player.PlayerInteractEvent.class, remap = false)
-public abstract class MixinEventPlayerInteractBlock extends PlayerEvent implements PlayerInteractBlockEvent {
+public abstract class MixinEventPlayerInteractBlock extends PlayerEvent implements PlayerInteractBlockEvent, IMixinEventPlayerInteractBlock {
 
     @Shadow
     public Action action;
@@ -58,6 +60,8 @@ public abstract class MixinEventPlayerInteractBlock extends PlayerEvent implemen
     public BlockPos pos;
 
     @Shadow public EnumFacing face;
+
+    public Optional<Vector3d> clickedPosition = Optional.absent();
 
     public MixinEventPlayerInteractBlock(EntityPlayer player, Action action, BlockPos pos, EnumFacing face, World world) {
         super(player);
@@ -92,4 +96,15 @@ public abstract class MixinEventPlayerInteractBlock extends PlayerEvent implemen
         return Optional.fromNullable(new Cause(null, this.entityPlayer, null));
     }
 
+    @Override
+    public void setClickedPosition(Vector3d clickedPosition) {
+        this.clickedPosition = Optional.of(VecHelper.toVector(this.pos).toDouble().add(clickedPosition));
+    }
+
+    @Override
+    public Optional<Vector3d> getClickedPosition() {
+        if (!clickedPosition.isPresent())
+            clickedPosition = Optional.of(VecHelper.toVector(this.pos).toDouble());
+        return clickedPosition;
+    }
 }
--- a/src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
@@ -103,8 +103,6 @@ public abstract class MixinEventPlayerInteractBlock extends PlayerEvent implemen
 
     @Override
     public Optional<Vector3d> getClickedPosition() {
-        if (!clickedPosition.isPresent())
-            clickedPosition = Optional.of(VecHelper.toVector(this.pos).toDouble());
         return clickedPosition;
     }
 }
--- a/src/main/java/org/spongepowered/mod/interfaces/IMixinEventPlayerInteractBlock.java
+++ b/src/main/java/org/spongepowered/mod/interfaces/IMixinEventPlayerInteractBlock.java
@@ -1,0 +1,31 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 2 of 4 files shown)


src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
```diff
@@ -79,4 +85,15 @@ public EntityInteractionType getInteractionType() {
         return Optional.fromNullable(new Cause(null, this.entityPlayer, null));
     }
 
+    @Override
+    public Optional<Vector3d> getClickedPosition() {
+        return Optional.of(VecHelper.toVector(this.pos).toDouble());
```
simon816: Possibly could store this in a field to avoid creating 3 objects every call

thomas15v: I actually copied this over from somewhere else. But indeed I agree with it.

Zidane:
@thomas15v 

Can you address your feedback and update this for latest changes?

thomas15v:
Not sure put shouldn't this be pulled in SpongeCommon?

simon816:
This is specifically mixing into a forge event so it should be here.

thomas15v:
Whoops sorry :worried:. The heat is killing me here. I update it :smile:.

src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
simon816: Looking at this again, `this.pos` is the block position, not the clicked position.
There is no field in the forge event for the clicked position, so you'll have to do an `@Inject` into `NetHandlerPlayServer.processPlayerBlockPlacement`
Specifically, the packet, `C08PacketPlayerBlockPlacement`, has 3 methods you need.
`getPlacedBlockOffsetX()`
`getPlacedBlockOffsetY()`
`getPlacedBlockOffsetZ()`

thomas15v: But than this mixin is kinda useless. As the event will have to be called from ```NetHandlerPlayServer.processPlayerBlockPlacement```. 

Isn't it possible to figure out the clicked position from the players position?

simon816: The offset is where the crosshair on the client is pointing on the block. It's relative to the block in its 3D space, so x,y,z all range from 0 to 1. (Outside that range is invalid and should return absent). From some testing, when x = 0, that's the most west of the block, similarly with z = 0 that's the most north of the block, and y is the height where 0 is the bottom and 1 is the top.

This mixin is by no means useless, you just need to add an additional method on the event to be able to set the position before posted to the event bus.
The method `ItemInWorldManager.activateBlockOrUseItem` is invoked as a result of the packet, its last 3 parameters (`float p_180236_6_, float p_180236_7_, float p_180236_8_`) are the x y and z offset directly from the packet. It's possible to capture these parameters using an inject. The inject will be just before the packet is posted, so that you can call the new method on the event to set these values.

thomas15v: The problem is that the event itself gets called from ```ForgeEventFactory```. If I modify the event object in ```MixinNetHandlerPlayServer```. The event will be called already. So that means I need to make a mixin for ```ForgeEventFactory```. But even that wouldn't work, as I haven't found anything in mixins to change/modify code.

thomas15v: Forget what I said above. I think I might have founded it. But ehm:
```java
@Redirect(method = "processPlayerBlockPlacement", at = @At(value = "INVOKE", target = "Lnet/minecraftforge/event/ForgeEventFactory;onPlayerInteract(Lnet/minecraft/entity/player/EntityPlayer;Lnet/minecraftforge/event/entity/player/PlayerInteractEvent$Action;Lnet/minecraft/world/World;Lnet/minecraft/util/BlockPos;Lnet/minecraft/util/EnumFacing;)Lnet/minecraftforge/event/entity/player/PlayerInteractEvent;"))
public PlayerInteractEvent onC08(EntityPlayer player, PlayerInteractEvent.Action action, World world, BlockPos pos, EnumFacing face, C08PacketPlayerBlockPlacement packetIn) {
    System.out.println(new javax.vecmath.Vector3d(packetIn.getPlacedBlockOffsetX(), packetIn.getPlacedBlockOffsetY(), packetIn.getPlacedBlockOffsetZ()));
    return ForgeEventFactory.onPlayerInteract(player, action, world, pos, face);
}
```
```
[12:35:23] [Server thread/INFO] [STDOUT]: [net.minecraft.network.NetHandlerPlayServer:onC08:88]: (0.0, 0.0, 0.0)
[12:35:27] [Server thread/INFO] [STDOUT]: [net.minecraft.network.NetHandlerPlayServer:onC08:88]: (0.0, 0.0, 0.0)
[12:35:27] [Server thread/INFO] [STDOUT]: [net.minecraft.network.NetHandlerPlayServer:onC08:88]: (0.0, 0.0, 0.0)
[12:35:28] [Server thread/INFO] [STDOUT]: [net.minecraft.network.NetHandlerPlayServer:onC08:88]: (0.0, 0.0, 0.0)
[12:35:28] [Server thread/INFO] [STDOUT]: [net.minecraft.network.NetHandlerPlayServer:onC08:88]: (0.0, 0.0, 0.0)
[12:35:28] [Server thread/INFO] [STDOUT]: [net.minecraft.network.NetHandlerPlayServer:onC08:88]: (0.0, 0.0, 0.0)
[12:35:29] [Server thread/INFO] [STDOUT]: [net.minecraft.network.NetHandlerPlayServer:onC08:88]: (0.0, 0.0, 0.0)
[12:35:29] [Server thread/INFO] [STDOUT]: [net.minecraft.network.NetHandlerPlayServer:onC08:88]: (0.0, 0.0, 0.0)
```
The data the client sends when that event occurs isn't very interesting.

simon816: Well it definitely worked when I tried it. `javax.vecmath.Vector3d` is the wrong class, use `com.flowpowered.math.vector.Vector3d`
But you are correct in using `@Redirect`, with redirect you can replace existing code which is what needs to be done.
When it comes to mixins sometimes it can be quite messy but there's almost always a way.

thomas15v:
The old commit got lost :worried:. So this PR only addresses the exact click position.

src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
```diff
@@ -92,4 +96,15 @@ public Direction getSide() {
...
 
+    @Override
+    public void setClickedPosition(Vector3d clickedPosition) {
+        this.clickedPosition = Optional.of(VecHelper.toVector(this.pos).toDouble().add(clickedPosition));
+    }
+
+    @Override
+    public Optional<Vector3d> getClickedPosition() {
+        if (!clickedPosition.isPresent())
+            clickedPosition = Optional.of(VecHelper.toVector(this.pos).toDouble());
```
simon816: This should remain absent if the position is not available. The block location is already known due to the event extending `BlockEvent`

thomas15v: Oh right.

src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
```diff
@@ -92,4 +96,13 @@ public Direction getSide() {
         return Optional.fromNullable(new Cause(null, this.entityPlayer, null));
     }
 
+    @Override
+    public void setClickedPosition(Vector3d clickedPosition) {
+        this.clickedPosition = Optional.of(VecHelper.toVector(this.pos).toDouble().add(clickedPosition));
```
simon816: Not sure if this should be relative to the block or to the world. According to [git blame](https://github.com/SpongePowered/SpongeAPI/commit/57d70e3ccbde63583c3ee58bbf5059a249ac8e84#diff-458e8e0f271d0d47c8a82eb451d9d538R57) gabizou created the method. Which should it be @gabizou?

Zidane:
New events change this drastically.

PR #284, opened 2015-05-28
Add constraints to @Overwrite methods in mixins

Mumfrey:
I'm pull-requesting this rather than committing directly because I'd like to get some feedback on the constraint ranges I've added here.

In a nutshell, constraints allow hard failures to be raised at both build and run time which have particular purposes:

* Forces us to keep on top of `@Overwrite` methods when bumping forge versions. Having a hard declaration on the annotation means that a contract is baked directly into the code which says "this overwrite is valid for this version and, say, the next 10 versions".

* Causes a hard crash when a user tries to run the coremod in a version of Forge which is just too far away from the version we built against for comfort. Basically we want to avoid the scenario where something is patched/fixed/improved in Forge, and our boistrous coremod comes along and just blithely overwrites the new stuff.

Thus the overwrite constraints aren't hard-and-fast, they're going to depend - on a per-method basis - on the anticipated volatility of the method in question, both in terms of how severe our changes are, and how often we anticipate the target method itself will change. Given we may know about the future (if working against an older forge version like we are right now), then we can even impose a hard upper bound on overwrites when we know the build number of a future breaking change.

Constraints use a pretty straightforward syntax, for a quick intro consult the [Constraint javadoc](http://jenkins.liteloader.com/job/Mixin/javadoc/org/spongepowered/asm/util/ConstraintParser.Constraint.html) which gives a bunch of examples.

For the constraints defined here, I've gone with a pretty straightforward approach of `1371+` (meaning 1371 and any later version) where I feel it's safe to do so. `1371 +10` for particularly hairy bits of code, and longer constraints varying between 50 and 100 for less critical bits of code which should nevertheless be periodically reviewed.

I suggest we make it a policy to never author a **new** constraint with a version lower than the version we're building against, but can feel free to let constraints trail a little bit when bumping forge versions depending on anticipated code volatility.

Welcome feedback and suggested adjustments to these values. I should mention that I [already committed](https://github.com/SpongePowered/Sponge/commit/d0e57450bdaa510c5c476bfc52f9dc7744d682b1) changes which provide a nice user-facing message when a constraint violation is encountered, which tells the user exactly what happens and suggests remedial actions.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/event/MixinEventBus.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/event/MixinEventBus.java
@@ -48,7 +48,7 @@ public abstract class MixinEventBus {
     @Shadow
     private IEventExceptionHandler exceptionHandler;
 
-    @Overwrite
+    @Overwrite(constraints = "FML(1371)")
     public boolean post(Event event) {
         IEventListener[] listeners = event.getListenerList().getListeners(this.busID);
 
--- a/src/main/java/org/spongepowered/mod/mixin/core/event/MixinEventBus.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/event/MixinEventBus.java
@@ -48,7 +48,7 @@ public abstract class MixinEventBus {
     @Shadow
     private IEventExceptionHandler exceptionHandler;
 
-    @Overwrite(constraints = "FML(1371)")
+    @Overwrite(constraints = "FML(1371 +10)")
     public boolean post(Event event) {
         IEventListener[] listeners = event.getListenerList().getListeners(this.busID);
 
--- a/src/main/java/org/spongepowered/mod/mixin/core/server/MixinServerConfigurationManager.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/server/MixinServerConfigurationManager.java
@@ -89,6 +89,8 @@ import java.util.List;
 @Mixin(ServerConfigurationManager.class)
 public abstract class MixinServerConfigurationManager implements IMixinServerConfigurationManager {
 
+    private static final String $OVERWRITE_POLICY = "FORGE(1371 +10)";
+
     @Shadow
     private static Logger logger;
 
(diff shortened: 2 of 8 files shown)


src/main/java/org/spongepowered/mod/mixin/core/server/MixinServerConfigurationManager.java
```diff
@@ -89,6 +89,8 @@
 @Mixin(ServerConfigurationManager.class)
 public abstract class MixinServerConfigurationManager implements IMixinServerConfigurationManager {
 
+    private static final String $OVERWRITE_POLICY = "FORGE(1371 +10)";
```
simon816: Is the `$` necessary? I don't like the use of it for a variable name

src/main/java/org/spongepowered/mod/mixin/core/event/MixinEventBus.java
```diff
@@ -48,7 +48,7 @@
     @Shadow
     private IEventExceptionHandler exceptionHandler;
 
-    @Overwrite
+    @Overwrite(constraints = "FML(1371 +10)")
```
simon816: I doubt the event bus will change much (at all) between forge builds.

src/main/java/org/spongepowered/mod/mixin/core/server/MixinServerConfigurationManager.java
Mumfrey: I like using $ to indicate that something is a structural member rather than something which should be read as part of the class

src/main/java/org/spongepowered/mod/mixin/core/event/MixinEventBus.java
Mumfrey: Yes but it's too scary a prospect to not mark it as a critical bit of infrastructure which should be monitored closely. Suggest a different tolerance?

Zidane:
I've remade this issue in SpongeCommon:

https://github.com/SpongePowered/SpongeCommon/issues/52

gabizou:
*waltzes in after almost a year passing by* So... is this going to happen?

Mumfrey:
I made a branch with constraints and asked for feedback, I never got any.

gabizou:
Well, I wouldn't mind adding constraints for both SF and SC and SV if it means that SF's overwrites are build dependent and SC's are MC version dependent. Not sure whether @Zidane is in agreement for 1.9.4's bleeding builds as those sometimes may have changes we may need to make and or update.

gabizou:
Please update the target branch to bleeding (Edit this PR and change the target branch) and reopen.

PR #302, opened 2015-06-28
Fix Potion Effects When Switching World

verbeto:
Self explanatory- fixes the bug where potion effects would not be updated on the client when they were teleported to another dimension. Requested to be changed by a pull request by @Zidane.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/entity/MixinEntity.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/entity/MixinEntity.java
@@ -111,6 +111,11 @@ public abstract class MixinEntity {
             entityplayermp1.theItemInWorldManager.setWorld(toWorld);
             entityplayermp1.addSelfToInternalCraftingInventory();
             entityplayermp1.setHealth(entityplayermp1.getHealth());
+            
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/mod/mixin/core/entity/MixinEntity.java
```diff
@@ -111,6 +111,11 @@ public boolean teleportEntity(net.minecraft.entity.Entity entity, Location locat
             entityplayermp1.theItemInWorldManager.setWorld(toWorld);
             entityplayermp1.addSelfToInternalCraftingInventory();
             entityplayermp1.setHealth(entityplayermp1.getHealth());
+            
+            for(Object effect : entityplayermp1.getActivePotionEffects()) {
+                entityplayermp1.playerNetServerHandler.sendPacket(new S1DPacketEntityEffect(entityplayermp1.getEntityId(), (PotionEffect) effect));
```
gabizou: Need the imports for this.

verbeto: Editing on the web is probably not the smartest idea! Imports fixed, should all be fine now.

PR #303, opened 2015-06-28
Fire Sponge events on the Forge bus

Aaron1011:
Currently, Forge events which implement Sponge interfaces are automatically fired on the Sponge event bus. This isn't done in reverse when a Sponge even is fired.

This PR fires Sponge events on the Forge bus by creating a new instance of a Forge event, which delegates any setter methods back to the Sponge event.

Only events created through the event generator are fired on the Forge bus. This is for two reasons:
* Complex situations can arise if an event implements multiple Sponge interfaces, when choosing which Forge events to instantiate. An event created from the event generator is a built-in event, which has a clearly defined function known to the event bus.
* Not firing other events, even if they only implement one Sponge interface, allows plugins to manually fire their event on the Forge bus, choosing exactly what Forge event(s) to create. Firing the event in all cases would prevent plugins from implementing their own logic without causing the event to be fired twice.

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeCoremod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeCoremod.java
@@ -40,26 +40,26 @@ import java.util.Map;
 
 @IFMLLoadingPlugin.MCVersion("1.8")
 public class SpongeCoremod implements IFMLLoadingPlugin {
-    
+
     public static final class TokenProvider implements IEnvironmentTokenProvider {
-        
+
         @Override
         public int getPriority() {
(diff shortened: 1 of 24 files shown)


simon816:
Perhaps the event mapping could be moved to [`EventRegistry`](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/event/EventRegistry.java) or move the event registry into  `SpongeModEventManager`?

Aaron1011:
@simon816: `EventRegistry` only exists to for FML state events, though even that could be done without it. I think this logic makes sense where it is.

PR #305, opened 2015-07-01
fix incorrect imports

aleshkanet:
manipulators => manipulator
entites => entity
items => item


RobertHerhold:
types => type

simon816:
I think the SDK should be added to the build path in the IDE, so that compile errors will be noticed immediately. All that needs to be added to the build.gradle file is
```grovy
sourceSets {
    sdk
}
```
@SpongePowered/developers thoughts?

PR #310, opened 2015-07-18
Implemented a preliminary LivingDeathEvent

DarkArc:
It's not complete, but it at least makes living death events fire in Sponge.

### DIFF
--- a/src/main/java/org/spongepowered/mod/event/SpongeModEventManager.java
+++ b/src/main/java/org/spongepowered/mod/event/SpongeModEventManager.java
@@ -52,6 +52,7 @@ import org.spongepowered.api.event.block.BlockUpdateEvent;
 import org.spongepowered.api.event.entity.EntityConstructingEvent;
 import org.spongepowered.api.event.entity.EntityEvent;
 import org.spongepowered.api.event.entity.EntitySpawnEvent;
+import org.spongepowered.api.event.entity.living.LivingDeathEvent;
 import org.spongepowered.api.event.entity.living.LivingEvent;
 import org.spongepowered.api.event.entity.player.PlayerBreakBlockEvent;
 import org.spongepowered.api.event.entity.player.PlayerChatEvent;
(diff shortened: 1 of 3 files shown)


src/main/java/org/spongepowered/mod/event/SpongeModEventManager.java
```diff
@@ -94,6 +95,7 @@
                     .put(BlockEvent.class, net.minecraftforge.event.world.BlockEvent.class)
                     .put(BlockUpdateEvent.class, net.minecraftforge.event.world.BlockEvent.NeighborNotifyEvent.class)
                     .put(LivingEvent.class, net.minecraftforge.event.entity.living.LivingEvent.class)
+                    // .put(LivingDeathEvent.class, net.minecraftforge.event.entity.living.LivingDeathEvent.class)
```
Aaron1011: Why are you commenting this out?

src/main/java/org/spongepowered/mod/mixin/core/event/entity/living/MixinEventLivingDeath.java
```diff
@@ -0,0 +1,52 @@
...
+        return getEntity().getLocation();
+    }
+
+    /*
+    private static net.minecraftforge.event.entity.living.LivingDeathEvent fromSpongeEvent(org.spongepowered.api.event.entity.living.LivingDeathEvent spongeEvent) {
+        LivingDeathEvent event = new LivingDeathEvent((EntityLivingBase) spongeEvent.getEntity(), Dam);
+        ((IMixinEvent) event).setSpongeEvent(spongeEvent);
+        return event;
+    }
+    */
```
DarkArc: Because I didn't finish this function yet...

DarkArc:
@Aaron1011 any suggestions on how to actually implement fromSpongeEvent, seeing as how the sponge event doesn't concern itself with damage causes, and the constructor for the forge version of the event requires a damage cause?

ryantheleach:
I'd change the API for the Sponge event to get the damage result/cause that resulted from the killing blow. Even though the death event shouldn't be cancellable in sponge, it would be useful to know what was the final cause.

DarkArc:
@ryantheleach right, but that's why the cause API exist. Though, I don't know what it's state is.

gabizou:
The cause API is also going through a change with SpongePowered/SpongeAPI#712 as well, but I agree, the causes is the correct place to store the direct cause for the event. If anything, the `DamageCause` can be recreated or vice versa.

Zidane:
Events API has changed drastically.

PR #383, opened 2015-10-24
Implement most SleepingEvents.

JBYoshi:
I couldn't implement `SleepingEvent.Post` with a Forge event mixin because it requires more data than `PlayerWakeUpEvent` gives, but I needed to use code more complex than `@Inject` allows, and Forge heavily modifies this method. `SleepingEvent.Tick` was implemented in SpongePowered/SpongeCommon#229.

Requires SpongePowered/SpongeCommon#283.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/entity/player/MixinEntityPlayer.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/entity/player/MixinEntityPlayer.java
@@ -24,21 +24,38 @@
  */
 package org.spongepowered.mod.mixin.core.entity.player;
 
+import org.spongepowered.common.Sponge;
 import org.spongepowered.common.mixin.core.entity.living.MixinEntityLivingBase;
+import org.spongepowered.common.util.VecHelper;
+
+import java.util.Optional;
 
 import org.spongepowered.asm.mixin.Shadow;
 import net.minecraft.entity.player.InventoryPlayer;
 import org.spongepowered.asm.mixin.Overwrite;
 import net.minecraft.entity.player.EntityPlayer;
+import org.spongepowered.api.entity.Transform;
+import org.spongepowered.api.event.SpongeEventFactory;
+import org.spongepowered.api.event.action.SleepingEvent;
+import org.spongepowered.api.event.cause.Cause;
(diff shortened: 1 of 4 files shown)


JBYoshi:
Closing and reopening to have Travis rebuild the PR - the latest build failed because of a "Connection reset".

JBYoshi:
Status?

src/main/java/org/spongepowered/mod/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -50,4 +67,48 @@ public EntityItem dropOneItem(boolean dropAll) {
     public EntityItem dropPlayerItemWithRandomChoice(ItemStack itemStackIn, boolean unused) {
         return this.dropItem(itemStackIn, false, false);
     }
+
+    @Overwrite
```
bloodmc: This should be moved to SpongeCommon. As for the PlayerWakeupEvent, just create a mixin to fire it before line 94.

JBYoshi: Actually, Forge changes more than just firing the event - it adds methods like `isBed()` and `setBedOccupied()`.

src/main/java/org/spongepowered/mod/mixin/core/entity/player/MixinEntityPlayer.java
```diff
@@ -24,21 +24,38 @@
...
 
 @Mixin(EntityPlayer.class)
 public abstract class MixinEntityPlayer extends MixinEntityLivingBase {
 
     @Shadow public InventoryPlayer inventory;
+    @Shadow public BlockPos playerLocation;
+    @Shadow protected boolean sleeping;
+    @Shadow private int sleepTimer;
     @Shadow public abstract EntityItem dropItem(ItemStack droppedItem, boolean dropAround, boolean traceItem);
+    @Shadow public abstract void setSize(float width, float height);
```
bloodmc: setSize is not part of the target class EntityPlayer but rather Entity so this needs to be removed.

JBYoshi:
Now requires SpongePowered/SpongeCommon#283.

PR #405, opened 2015-11-08
World Generation Phase 1 - Populators

### LINKED ISSUE #429: Get Custom Biomes from Mods
Hey guys,

is there at the moment the option to get custom biomes out of an other mod e.g. biomes o plenty to use it in the world generator?

Deamon5550:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/734) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/253) | SpongeForge

Closes #429

Known Issues / TODO:
- [x] biome tree type overrides in forge
- [x] validate that modded chunk providers are properly wrapped
- [x] Move deferred biome decorator event call to BiomeGenBase#initPopulators which will call it later (almost exactly at the time it would normally be called)
- [x] Call forge generation events from populators


src/main/java/org/spongepowered/mod/mixin/core/world/MixinWorld.java
```diff
@@ -44,5 +44,4 @@
     private void onUpdateWeatherBody(CallbackInfo ci) {
         this.setWeatherStartTime(this.worldInfo.getWorldTotalTime());
     }
-
```
Deamon5550: random new line is random

src/main/java/org/spongepowered/mod/event/SpongeEventHooks.java
```diff
@@ -24,10 +24,11 @@
  */
 package org.spongepowered.mod.event;
 
+import org.spongepowered.common.interfaces.entity.IMixinEntity;
```
Deamon5550: NTS these imports are probably not needed

PR #414, opened 2015-11-17
[WIP] Phase 1 - AI Framework

Zidane:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/920) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/264) | SpongeForge

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeMod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeMod.java
@@ -24,6 +24,7 @@
  */
 package org.spongepowered.mod;
 
+import com.google.common.base.Preconditions;
 import com.google.common.collect.ImmutableMap;
 import com.google.common.eventbus.EventBus;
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/mod/SpongeMod.java
```diff
@@ -143,6 +141,8 @@ public void onPreInit(FMLPreInitializationEvent event) {
             SpongeForgeModuleRegistry.registerForgeData();
             SpongeModMessageHandler.init();
 
+            Class<?> clazz = Class.forName("org.spongepowered.api.entity.ai.task.AbstractAITask");
```
Minecrell: Is this really needed?

Aaron1011: Not really, but it causes the transformer to run earlier. If there's any sort of problem , it will show up earlier rather than later.

Issue #404, opened 2015-11-07
Version conflict with com.typesafe:config in development environment


JBYoshi:
When updating dependencies, Forge and Sponge both have dependencies on `com.typesafe:config` with conflicting versions. Eclipse adds both to the classpath and gives priority to the Forge-provided version, resulting in errors at runtime.

Full console log up to the point the first exception was thrown:
```
[11:10:26] [main/INFO] [GradleStart]: userProperties: []
[11:10:26] [main/INFO] [GradleStart]: accessToken: NotValid
[11:10:26] [main/INFO] [GradleStart]: version: MCDev
[11:10:26] [main/INFO] [GradleStart]: username: JBYoshi
[11:10:26] [main/INFO] [GradleStart]: Extra: [--tweakClass, org.spongepowered.asm.launch.MixinTweaker]
[11:10:26] [main/INFO] [GradleStart]: Running with arguments: [--userProperties, [], --assetsDir, C:/Users/Jonathan/.gradle/caches/minecraft/assets, --assetIndex, 1.8, --accessToken{REDACTED}, --version, MCDev, --username, JBYoshi, --tweakClass, net.minecraftforge.fml.common.launcher.FMLTweaker, --tweakClass, org.spongepowered.asm.launch.MixinTweaker, --tweakClass, net.minecraftforge.gradle.tweakers.CoremodTweaker]
[11:10:26] [main/INFO] [LaunchWrapper]: Loading tweak class name net.minecraftforge.fml.common.launcher.FMLTweaker
[11:10:26] [main/INFO] [LaunchWrapper]: Using primary tweak class name net.minecraftforge.fml.common.launcher.FMLTweaker
[11:10:26] [main/INFO] [LaunchWrapper]: Loading tweak class name org.spongepowered.asm.launch.MixinTweaker
[11:10:26] [main/INFO] [mixin]: SpongePowered MIXIN Subsystem Version=0.4.6 Source=file:/C:/Users/Jonathan/.gradle/caches/modules-2/files-2.1/org.spongepowered/mixin/0.4.6-SNAPSHOT/6946b6fb004a2ebdaab7c5822736bd08d8cec308/mixin-0.4.6-SNAPSHOT.jar Env=CLIENT
[11:10:27] [main/INFO] [LaunchWrapper]: Loading tweak class name net.minecraftforge.gradle.tweakers.CoremodTweaker
[...]

PR #442, opened 2015-12-12
Exclude config-1.2.1 from eclipse classpath

kashike:
This fixes https://github.com/SpongePowered/SpongeForge/issues/404

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -79,6 +79,12 @@ shadowJar {
     }
 }
 
+eclipse.classpath.file.whenMerged {classpath ->
+    classpath.entries.removeAll(classpath.entries.findAll {
+        entry -> entry.kind == 'lib' && entry.path.contains('com.typesafe/config/1.2.1/') && entry.path.endsWith('config-1.2.1.jar')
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -79,6 +79,12 @@ shadowJar {
     }
 }
 
+eclipse.classpath.file.whenMerged {classpath ->
+    classpath.entries.removeAll(classpath.entries.findAll {
+        entry -> entry.kind == 'lib' && entry.path.endsWith('com.typesafe/config/1.2.1/f771f71fdae3df231bcd54d5ca2d57f0bf93f467/config-1.2.1.jar')
```
gabizou: And if this changes at all? How about something more generic where it's just a default search of `.contains('com.typesafe/config/1.2.1/')`?

kashike: Why would a released version's hash change?

gabizou: Surefire way to know that regardless of the hash that it's going to be excluded.

JBYoshi:
Why was this closed?

Issue #263, opened 2015-05-04
Ongoing Minor Issue List


Deamon5550:
__Minor Issues List__

- [ ] Discover Minor Issues

__Checkstyle warnings__

File | Problem
---- | ----
Unqualified access to instance field | [MixinEventPlayerChat:74](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerChat.java#L74)
Unqualified access to instance field | [SpongeModGameRegistry:103-104](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/registry/SpongeModGameRegistry.java#L103-L104)
Statement unnecessarily nested within else clause | [CommandSponge:258](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/command/CommandSponge.java#L258)
Statement unnecessarily nested within else clause | [MixinEventBus:57](https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/mixin/core/event/MixinEventBus.java#L57)

liach:
[`SpongeMod.java`] (https://github.com/SpongePowered/Sponge/blob/master/src/main/java/org/spongepowered/mod/SpongeMod.java) does not need to implement `PluginContainer`

JBYoshi:
- [ ] Fix newlines at ends of files in 653f40bb749567ebcbc2c53fa4140a9bb82184b1

oliverdunk:
Newlines fixed in PR #467

randombyte-developer:
https://github.com/SpongePowered/SpongeForge/blob/7e578657914f7c5605ceeac548cf0c0fae46dd83/src/main/java/org/spongepowered/mod/event/SpongeForgeEventFactory.java#L858 typo

liach:
https://github.com/SpongePowered/SpongeForge/commit/e37fa119fa73a9a2819ad68a0d9ff8bb436f0459#diff-58e0014cf676b44128730ca938af8f0dR135
Should be `SERVER_THREAD` instead.

LordRalex:
https://github.com/SpongePowered/SpongeForge/blob/6d0ee2e4e438b86a3e7051bbfea16fca16554eea/src/main/java/org/spongepowered/mod/mixin/core/fml/common/MixinLoader.java#L90

[...]

PR #467, opened 2015-12-29
Adds newlines to files introduced in 653f40b

oliverdunk:
OCD fix.


kashike:
This should be added to https://github.com/SpongePowered/SpongeForge/issues/263 to be merged at a later time in bulk.

PR #473, opened 2015-12-31
FakePlayer changes for SpongeForge

simon816:
[SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/86) | **SpongeForge**

- Use Forge's FakePlayer as the simulated player
- Handle fake players in interact event


src/main/java/org/spongepowered/mod/SpongeModGame.java
```diff
@@ -85,4 +90,15 @@ public SpongeNetworkManager getChannelRegistrar() {
         return this.networkManager;
     }
 
+    @Override
+    public FakePlayer.Factory createFakePlayerFactory() {
+        return new FakePlayer.Factory() {
```
JBYoshi, line 74: I believe this could be replaced with `return FakePlayerFactory::get;`

simon816, line 74: Doesn't work because `FakePlayer.Factory` is not a functional interface

gabizou:
Please update the target branch to bleeding (Edit this PR and change the target branch) and reopen.

simon816:
updated

ryantheleach:
@simon816 @gabizou This issue is pretty old, and it's been updated a couple of times.

Does it keep getting missed, or apart from being updated, is there something that is legitimately blocking this?

simon816:
@ryantheleach Waiting on SpongePowered/SpongeCommon#86

ImMorpheus:
Closing this given the age that has passed and the lack of activity. If someone wants to continue the work done here then we can re-open at a later time.

PR #481, opened 2016-01-03
Update README.md

sync667:
Beta stage, that's right?


Zidane:
I'll leave it as-is. Thanks for the PR. I'll change it later to something a bit better when we have more stability.

PR #486, opened 2016-01-05
[WIP] Refactor explosion events

kashike:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1011) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/401) | [**Forge**](https://github.com/SpongePowered/SpongeForge/pull/486)


src/main/java/org/spongepowered/mod/event/SpongeForgeEventFactory.java
```diff
@@ -683,17 +683,17 @@ public static ChunkEvent createChunkEvent(Event event) {
         }
 
         net.minecraft.world.World forgeWorld = (net.minecraft.world.World) world.get();
-        Explosion explosion = (Explosion) spongeEvent.getExplosion();
+        Explosion explosion = (Explosion) null;//spongeEvent.getExplosion();
```
Zidane: Why was this set to null?

kashike: These PRs are still in-progress

gabizou: Then mark the PR's as such? I don't see a `[WIP]` in the title of any of them.

kashike: I'd actually forgotten about this until now, I had meant to comment on it in the description - the `Explosion` is not available in the Sponge `ExplosionEvent` parent, only `Start`, `Detonate`, and `Post`. I'm not sure what to do here other than:
- passing `null` to the Forge event constructor
- throwing an exception if this method is called
 - Forge does not call this method natively, mods might, so:
- printing a warning to console and returning `null`

kashike:
Superseded by https://github.com/SpongePowered/SpongeForge/pull/513

PR #505, opened 2016-01-19
Add ItemStack to cause for interact events.

ryantheleach:
Adds ItemStacks to interact events. Tested with a simple listener listening to all interact events printing causes to console and logging.

Pretty sure I've gotten this right, not sure what to do about SpongeVanilla or how to correctly commit SpongeCommon yet.


ryantheleach:
And this one is targetted at master :(

PR #506, opened 2016-01-19
Add ItemStacks to interact events.

ryantheleach:
This commit adds ItemStack Causes to Interact Events fired by the player.

The reasoning behind this, is by using this you can now filter based on itemstack, as well as preparing for mods / 1.9 where the slot that is causing the interact may not be the players currently held item.

Will probably require Vanilla to implement the equivalent of the forge changes, as well as a way to catch modded interactions before this is finished.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
@@ -37,6 +37,7 @@ import org.spongepowered.api.block.BlockTypes;
 import org.spongepowered.api.event.Event;
 import org.spongepowered.api.event.SpongeEventFactory;
 import org.spongepowered.api.event.block.InteractBlockEvent;
+import org.spongepowered.api.event.cause.Cause;
 import org.spongepowered.api.util.Direction;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.api.world.Location;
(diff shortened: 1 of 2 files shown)


src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractBlock.java
```diff
@@ -110,12 +110,16 @@ public void syncDataToForge(Event spongeEvent) {
 
     @Override
     public Event createSpongeEvent() {
+        Cause cause = getCause();
+        if(StaticMixinHelper.prePacketProcessItem != null){
+            cause = cause.with(StaticMixinHelper.prePacketProcessItem);
```
gabizou: You should give this a useful name.

src/main/java/org/spongepowered/mod/mixin/core/event/player/MixinEventPlayerInteractEntity.java
```diff
@@ -52,6 +53,9 @@
     @Inject(method = "<init>", at = @At("RETURN"))
     private void onConstruct(CallbackInfo callbackInfo) {
         this.cause = Cause.of(NamedCause.source(((EntityEvent) (Object) this).entity));
+        if(StaticMixinHelper.prePacketProcessItem != null){
+            this.cause = this.cause.with(StaticMixinHelper.prePacketProcessItem);
```
gabizou: And here.

Aaron1011:
@ryantheleach: This should really target `master`, as there's no corresponding API change in `bleeding`.

Zidane:
@ryantheleach 

I'm closing this solely to have you do it in Common instead please.

ryantheleach:
Sorry I had forgotten that I had this pending.
@gabizou Is this still needed with the current cause refactoring?

gabizou:
Not likely, a majority of contextual data for events will be ironed out with @Zidane's cause refactor.

PR #519, opened 2016-01-31
Update the SDK

simon816:
Update to ForgeGradle 2
Update reference to SpongeForge instead of Sponge
Provide simple plugin template rather than a more complete plugin


sdk/build.gradle
```diff
@@ -1,36 +1,48 @@
...
 repositories {
+    mavenCentral()
+    maven {
+        name = "sponge"
+        url = "http://repo.spongepowered.org/maven"
+    }
     maven {
-        name "sponge"
-        url "http://repo.spongepowered.org/maven/"
+        name "spongeforge"
```
kashike: `name "forge"`

Minecrell: Also use `=` for consistency with above

sdk/build.gradle
```diff
@@ -1,36 +1,48 @@
...
             url = "http://files.minecraftforge.net/maven"
         }
-        maven {
-            name = "sonatype"
-            url = "https://oss.sonatype.org/content/repositories/snapshots/"
-        }
     }
     dependencies {
-        classpath 'net.minecraftforge.gradle:ForgeGradle:1.2-SNAPSHOT'
+        classpath 'net.minecraftforge.gradle:ForgeGradle:2.1-20160116.192356-17'
```
liach: Why not just `2.1-SNAPSHOT`?

simon816: I used the exact version SpongeForge was on at the time

Minecrell:
I'm going to close this, once I find the time I'll start on getting something into SpongeGradle which will allow for easier testing with both implementations.

PR #529, opened 2016-02-09
Update Flardians example for new Text and Data APIs

ZephireNZ:
> FLARD. As it says in the Pirate Translations; leave it alone and it'll probably leave you alone.

Maybe that's why the SDK example hasn't been updated for the Beta API changes?

This fixes references to the old `Texts.of`, as well as using the new `DataManager` system.


Minecrell:
The SDK has been updated in #519 where the plugin was also replaced with a simpler one because that is really just supposed to be used as a base. The updated Flardians is already available in the Cookbook. Thank you for your contribution, however the SDK update is going to be merged soon.

PR #572, opened 2016-03-14
Fix mods using null to remove items after use

modwizcode:
This solves an issue with the vials in Botania causing infinite duplication after their last use. The mod handles storing empty vials in a custom way and then returns null after the event to delete the item. This breaks the code because results are supposed to be non-null so snapshots can be created.

To solve this problem, I made the null item case be replaced with a copy that has a 0 stack size. Which should remove the item, while not requiring significant modifications to the way the snapshot transactions work.


kashike:
The `result` field can be changed by listeners of the `PlayerUseItemEvent.Finish` event.

modwizcode:
@kashike I fail to see the importance of this? If they set it to null that should be considered the same as mod behavior which is to remove the item. That is what happens under forge, the behavior under SpongeForge currently is crashing.

modwizcode:
Ah ok. I see, they could unset the item for instance. I don't see how that's any different from if this mod just set stackSize to 0 like vanilla does to get rid of a used item such as food. The behavior should be the same. Which means listeners have the same ability to change it as they do with food.

PR #609, opened 2016-04-14
[WIP] Refactor BungeeCord support to more generic 'Proxy' support

kashike:
[Common](https://github.com/SpongePowered/SpongeCommon/pull/649) | [**Forge**](https://github.com/SpongePowered/SpongeForge/pull/609)


bloodmc:
@dualspiral can you take a look at this PR when you get a chance? Thanks!

dualspiral:
Thinking about this, I'm beginning to wonder if there is a good reason for all this abstraction. Normally, I'm all for it, but in this case, I'm not sure there is.

That aside for the moment, the problems I have right now (I'm at work so haven't been able to test this or look in depth, these comments also cover https://github.com/SpongePowered/SpongeCommon/issues/649)

* `IMixinCHandshake` should be `IMixinC00Handshake`.
* Methods `isFailed` and `setFailed` interact with a variable `cancelled` in the `HandshakeResponse` class - shouldn't it be `failed` for consistency?
* There is no way to specify _which_ proxy implementation you want, only that you want _a_ proxy implementation. This abstraction is useless if you can't use it to switch implementations out. I know there is only Bungee at the moment, but the whole point of this is to make it easier to add proxy implementations. So, how would you do that?
* Renaming "IP Forwarding" to "Forward Client Details" might not be a smart idea. The general server owner knows it as "IP forwarding", and while I know the latter is a much truer representation of what it is, you _are_ going to get admins who just start asking about the option.

The original complaint was that we constrained the proxy implementation to Bungee only. I agree, other proxies should be supported. However, you're doing a similar thing with this abstraction, it feels like you are unintentionally saying that if you want to add an implementation, it has to be done using this framework. Personally, I'd actually _want_ to see a constrained and secure specification that proxies have to follow, but it certainly wouldn't be Bungee's way of doing it, and I think we've have a hard time trying to convince the Bungee folks to change their ways.

Consider a different scenario: say I decide to alter (or create) a proxy to have a new forwarding protocol. I introduce a new login phase with an ID of -1, which then interacts with the proxy using a custom set of packets to perform a more secure login phase than just sending extra data to the handshake and hoping the firewall works. Now, all that's changed in my handshake packet is that it's a new ID - this abstraction would be pretty much useless to me, and I'd need a full new mixin plugin to do it anyway.

Technically, at a quick glance, I don't think there is anything wrong with this, so if you decide to continue with this, then I'll support it and pitch in if/when I can. I really hate to see effort go to waste. However, because it was originally designed to work with the API, this feels like this is an abstraction for an API that doesn't exist, and that implementations would be better served just being their own as small as possible mixin plugins (`proxy.waterfall`, `proxy.bungeecord`, `proxy.lillypad` etc.). I agree that some of the Bungee code here could do with a cleanup, but it feels to me that all that's been done is that about 700 lines of code have been added to allow you to read a modified handshake with more constraints than before, with an internal API that some implementations might not use anyway.

@bloodmc I'll test this more when I get chance to do so at home if you wish.

@gabizou @Zidane - you're probably interested in these comments too.

kashike:
@dualspiral This wasn't written for the API, actually, it was freshly written to be in SpongeCommon/SpongeForge only - if plugins **do** decide to set their own proxy, they can:
```java
Proxy myProxy = ...;
ProxyManager.INSTANCE.setProxy(myProxy);
```

The "BungeeCord" format has become the standard for LilyPad as well, and potentially other solutions. If you've any idea how to not make it so reliant on the handshake information, please comment.

kashike:
I should also mention that you aren't **required** to modify the handshake, your implementation can obtain details using other methods (database? cache? who knows).

dualspiral:
> if plugins do decide to set their own proxy, they can

Plugins from where? Separate plugins that depend on SpongeCommon? **If that's the case**, my original comment still stands that plugins should not be allowed to easily disrupt the login process. I would _strongly_ recommend that the _only_ way that implementations can set client details is through a mixin plugin which is reviewed by the Sponge team and any requested changes go through them.

**It should not be possible for a Sponge plugin** (depending on whatever) **to simply replace the instance** - which is what I think is going on here. This is a security problem, even if you have to activate the mixin first, but there are enough server networks out there that _will_ be using networks. Admins should be explicitly stating how they want users to authenticate, without some rouge plugin being able to change it.

Even if the access modifier is set low enough, having such abstraction in place still makes it easy for me to "break in" and update the credentials - so if the plugin detects me coming online, it'll swap my profile for yours, for example. I won't say that the Bungee plugin as it stands is perfect, but it's less vulnerable to such an attack.

_The rest of this comment is more on how I think this fowarding should be done, and why I don't think that the handshake mutation is the right way to go._

> If you've any idea how to not make it so reliant on the handshake information, please comment.

I believe that the way Bungee does it is insecure and should not be relied upon, and some of these concerns also stand in the present (currently committed) form of this client detail forwarding. Using the handshake packet is just plain insecure, and it feels like this is the path of least resistance. Server owners then need to setup firewalls - and believe me, there are a load of them who don't, I could quite easily assume the admin's identity and take them on.

There really should be a chain of trust between proxy and server. IP Forwarding does not have that, except for those who use a firewall properly. I would much rather see a different login phase where custom packets are exchanged during the login phase, including some way to authenticate the proxy to the server (public-private key stuff?), before sending the profile and such to the server.

> I should also mention that you aren't required to modify the handshake, your implementation can obtain details using other methods (database? cache? who knows).

It shouldn't have to be done so early - my argument is that it should be done during the login phase.

kashike:
> Plugins from where? Separate plugins that depend on SpongeCommon? If that's the case, my original comment still stands that plugins should not be allowed to easily disrupt the login process. I would strongly recommend that the only way that implementations can set client details is through a mixin plugin which is reviewed by the Sponge team and any requested changes go through them.

Yes, plugins that depend on SpongeCommon.

> I would strongly recommend that the only way that implementations can set client details is through a mixin plugin which is reviewed by the Sponge team and any requested changes go through them.

This is the one **big** issue with Spigot/etc - projects like LilyPad end up reflecting everything because they can't add a new proxy provider.

> It shouldn't have to be done so early - my argument is that it should be done during the login phase.

...and you would suggest what? If a proxy determines that a client isn't what it is expecting (public-private key, etc), the client **should not** reach past the handshaking process. It's also required in handshake because the client IP isn't the true client IP - this is why client detail forwarding came into existence.
___
Things won't start to change in the proxy world until something better replaces BungeeCord, or BungeeCord finally starts making some changes to it's authentication protocol - which I can't see being done anytime soon, they've yet to pull a PR that adds Forge/Sponge  support.

dualspiral:
> Yes, plugins that depend on SpongeCommon.

**Then I recommend rejection of this PR based on what was said in the API PR** - it's the same problem. Plugins **SHOULD NOT** be able to handle the login process. We went through this on the API PR, only Sponge should be determining what's good and what's not, not a plugin.

> This is the one big issue with Spigot/etc - projects like LilyPad end up reflecting everything because they can't add a new proxy provider.

@gabizou said that there is no problem with others reaching out to provide a mixin plugin to do it in the API PR. These things should be reviewed by the Sponge team, not be left to plugins. This argument has been made before.

> ...and you would suggest what?

Exactly as I said. The only change would be you would have to add a switch to a new protocol with a different ID, say "3", then use a series of packets to set the data.

> Things won't start to change in the proxy world until something better replaces BungeeCord, or BungeeCord finally starts making some changes to it's authentication protocol - which I can't see being done anytime soon, they've yet to pull a PR that adds Forge/Sponge support.

Yep, it requires a joint effort, but it should be done in a way that doesn't break the chain of trust between client and server, and a plugin can very well do that.

kashike:
Well, good luck to you, @dualspiral, on getting Spigot, BungeeCord, LilyPad, Forge, etc to agree on something.

gabizou:
@kashike so is it virtually impossible to clean up the proxy implementation to allow at least some sort of abstraction layer that allows server owners to choose which proxy implementation to use? I mean, feel free to include some sort of proxy implementations already existing, such as LilyPad etc.

PR #712, opened 2016-06-26
Implemented Player chat simulation method. Adds SpongeAPI-#733

me4502:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1283) | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/253) | [Forge](https://github.com/SpongePowered/SpongeForge/pull/712)

This PR adds a method to Player, that allows for the simulation of chat.

The exact reasoning of it can be seen in the corresponding issue, https://github.com/SpongePowered/SpongeAPI/issues/733

This method returns the event so that the plugin can see what modifications have been made by other plugins, and to see if it has been cancelled.

This has not been implemented in Common as SV/SF already has all chat handling separated. To put it in common would require moving all of the handling over to common.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/entity/player/MixinEntityPlayerMP.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/entity/player/MixinEntityPlayerMP.java
@@ -52,7 +52,7 @@ public abstract class MixinEntityPlayerMP extends MixinEntityPlayer implements P
     }
 
     @Override
-    public MessageChannelEvent.Chat chat(Text message) {
+    public MessageChannelEvent.Chat simulateChat(Text message) {
         String messageRaw = SpongeTexts.toLegacy(message);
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/mod/mixin/core/entity/player/MixinEntityPlayerMP.java
```diff
@@ -26,16 +26,49 @@
...
+        EntityPlayerMP thisPlayer = (EntityPlayerMP)(Object) this;
+
+        final ServerChatEvent event = new ServerChatEvent(thisPlayer, messageRaw, itextcomponent);
+        ((IMixinInitCause) event).initCause(Cause.of(NamedCause.source(this)));
+
+        if (!MinecraftForge.EVENT_BUS.post(event)) {
+            MessageChannelEvent.Chat spongeEvent = (MessageChannelEvent.Chat) event;
+            Text theMessage = spongeEvent.getMessage();
+            if (!spongeEvent.isMessageCancelled()) {
+                spongeEvent.getChannel().ifPresent(channel -> channel.send(this, theMessage, ChatTypes.CHAT));
```
kashike: You can just use `spongeEvent.getMessage()` instead of creating a local variable for it

PR #729, opened 2016-07-05
Implement ZombieData for Forge registries

ZephireNZ:
Common: SpongePowered/SpongeCommon#826
API: SpongePowered/SpongeAPI#1299

This changes the way the data processors work to use Forge's VillagerRegistry.

A note: When a vanilla profession is set on a zombie, NMS `ZombieType` will correspond to the correct profession name. However, when a mod profession is used, forge instead uses the `NORMAL` type - this also fixes this behaviour so that a zombie with a modded profession will still show as a `VILLAGER` zombie.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/entity/living/MixinEntityZombie.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/entity/living/MixinEntityZombie.java
@@ -1,0 +1,63 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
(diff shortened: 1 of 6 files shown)


bloodmc:
@ZephireNZ please update this PR and make sure it compiles.

src/main/java/org/spongepowered/mod/mixin/core/entity/living/MixinEntityZombie.java
```diff
@@ -0,0 +1,64 @@
...
+@Mixin(value = EntityZombie.class, priority = 1001, remap = false)
+public abstract class MixinEntityZombie implements Zombie {
+
+    @Shadow @Nullable public abstract VillagerRegistry.VillagerProfession getVillagerTypeForge();
+    @Shadow @Nullable public abstract net.minecraft.entity.monster.ZombieType getZombieType();
+
+    // Changes method on Common mixin, to use Forge professions
+    @Override
+    public ZombieData getZombieData() {
+        System.out.println("Forge zombie");
```
Aaron1011: Can you remove this?

ZephireNZ:
@gabizou Was there a reason this wasn't merged alongside the other two PRs?

ZephireNZ:
*ahem*

PR #794, opened 2016-07-27
Fix Dungeon mobs for Forge registry

ZephireNZ:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1226) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/710) | **Forge**

This adds a hook to forge's `DungeonHooks` registry so that the full list of `EntityArchetype`s that could be used by a `Dungeon` can be requested by plugins.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/forge/MixinDungeonHooks.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/forge/MixinDungeonHooks.java
@@ -1,0 +1,67 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/mod/mixin/core/forge/MixinDungeonHooks.java
```diff
@@ -0,0 +1,70 @@
...
+import org.spongepowered.common.entity.EntityUtil;
+import org.spongepowered.common.mixin.core.world.gen.populators.MixinWorldGenDungeons;
+
+import java.util.Optional;
+
+@Mixin(DungeonHooks.class)
+public class MixinDungeonHooks {
+
+    static {
+        MixinWorldGenDungeons.DEFAULT_ENTITIES_INTERNAL.clear();
```
gabizou: This will not work in runtime, mixin classes aren't available.

ZephireNZ: I couldn't remember if that was the case.

Would you recommend setting up something like DungeonMobRegistryModule? Or having a helper of some sort in SpawnerUtils?

gabizou: Module works.

bloodmc:
@ZephireNZ please update this PR and make sure it compiles.

ZephireNZ:
I haven't based on the latest bleeding because it breaks the build when using it with the `feature/schematics` branch on SpongeCommon - which I've based my spawner API upon due to the addition of archetypes.

PR #816, opened 2016-08-06
Implement MountEntityEvent and DismountEntityEvent.

JBYoshi:


bloodmc:
@JBYoshi This should be implemented in common. If we have no event for these in our API then they should be added. I'm trying to do away with having forge only events.

me4502:
@bloodmc There's already a PR for adding this implementation in Common, https://github.com/SpongePowered/SpongeCommon/pull/765

Isn't this PR required anyway, to add compatibility with Forge's event?

PR #964, opened 2016-10-16
Fix incorrect unit test directory in Gradle project and failing test

ghost:
#### As discussed in IRC:

Gradle/Maven expects unit tests to be found at `src/test/java`, however SpongeForge has been using `src/main/test/java`. This is incorrect.

---

Appended class with `@PowerMockIgnore` annotation to prevent PowerMockito from mocking classes we don't want to mock.

Also adds another `EventManager#registerListeners()` call in `testSpawn()`, as this **_should_** be expected behavior (we don't want to fire an event if nothing is listening for it - wasted processing time).

### DIFF
--- a/src/test/java/org.spongepowered.mod/event/SpongeModEventManagerTest.java
+++ b/src/test/java/org.spongepowered.mod/event/SpongeModEventManagerTest.java
@@ -1,8 +1,29 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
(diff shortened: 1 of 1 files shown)


jamierocks:
The directory is `org.spongepowered.mod` it should be `org/spongepowered/mod`.

ghost:
Oops. Looks like someone who created the test before me had this as the directory name, and my commit just continues on that mistake. Never mind; I'll fix.

src/test/java/org.spongepowered.mod/event/SpongeModEventManagerTest.java
```diff
@@ -75,6 +96,7 @@ public void testSpawn() throws ClassNotFoundException {
 
         eventBus.register(forgeListener);
 
+        this.eventManager.registerListeners(this.plugin, listener);
```
JBYoshi, line 99: I don't think this is supposed to be here. This section of the test is making sure that the fields are `true` when only Forge listeners are registered.

JBYoshi (changes requested):
I was looking at this earlier. I don't think this test works any more. Most of the test can actually be moved to SpongeCommon; only the Forge event section needs to stay here.

**EDIT:** The Forge-independent part is already in SpongeCommon (`ShouldFireTest`).

ghost:
Quite possibly not. Been well over a year since I submitted this merge request. I'm still interested in supporting it. If I get time, I'll update the branch - has since been deleted from my local repo.

src/test/java/org.spongepowered.mod/event/SpongeModEventManagerTest.java
ghost: Original MR message stated

    Also adds another EventManager#registerListeners() call in testSpawn(), 
    as this should be expected behavior (we don't want to fire an event if
    nothing is listening for it - wasted processing time)."

You could be right, though. Could we have someone else's thoughts?

ghost:
See #1896 for an update, since I deleted the original repo in which this MR was from

PR #1089, opened 2016-12-07
Utilise GitHub templates

kashike:
[Common](https://github.com/SpongePowered/SpongeCommon/pull/1059) | [**Forge**](https://github.com/SpongePowered/SpongeForge/pull/1089)


ST-DDT:
Maybe we should add a phrase like.

> * [ ] I used the latest version
> * [ ] I could not use the latest version because of:_____________

or as plain comment.

> If you don't use the latest version , please provide a reason.

We should also do something similar for SpongeCommon. And maybe provide some help where to post an issue.

Aaron1011 (approved):

PR #1106, opened 2016-12-12
Fixed plugins name could be null

Zartec:
Plugin name could be `null` if the Plugin is not built via SpongeGradle.

This is causing exceptions in commands like `/sponge plugins` and the Timings, which are both using `plugin.getName()`.


kashike (approved):

stephan-gh:
To be honest, you should always have a plugin name and plugin version unless you are running in development environment. If you don't use SpongeGradle you have to set them manually.

I think it would be a better solution to throw an error if the `mcmod.info` does not contain a plugin name and/or version.

Zartec:
The `name` is not forced by the annotation as you can see [here](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/plugin/Plugin.java#L67)
If i don't enter a specific `name` in the annotation the processor falls back to the `id`.

So i'm not forced to set a `name` if i want to have the same `name` and `id`.
For the `mcmod.info` it should be the same.

Also the same thing is already done for SpongeVanilla: [here](https://github.com/SpongePowered/SpongeVanilla/blob/bleeding/src/main/java/org/spongepowered/server/plugin/MetaPluginContainer.java#L60)

kashike:
:+1: for this

stephan-gh (approved):

PR #1188, opened 2017-01-14
Generalize recipes and implement them

Limeth:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1449)|[SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1119)|*SpongeForge*

This is a WIP continuation of [kashike's PR](https://github.com/SpongePowered/SpongeAPI/pull/1098), with his consent.
It aims to generalize how recipes are registered and handled. Instead of checking the equality of the ingredients, it works by testing ingredient predicates. It also aims to let the user implement the recipe interfaces by themselves, making it, for example, possible to transfer NBT data from the ingredients to the result.
In this PR, `ItemStackSnapshot`s are used extensively. This is to prevent the user from modifying the original `ItemStack`. I could instead provide an `ItemStack#copy`, but that doesn't make it visually obvious the original item won't change.


Zidane:
Closing as this has been implemented. Thanks again @Limeth for your work on this, it was much appreciated.

PR #1248, opened 2017-02-06
Implemented getWorld into the LoadingTicket class

adpking:
The SpongeForge implementation currently does not provide a way for the ticket to provide the world from the ticket, despite having provided it during the creation of the ticket. Forge supports this, so this is essentially just adding to the completion of the Forge Chunk Manager wrapper.

This is paired with my SpongeAPI pull request [#1479](https://github.com/SpongePowered/SpongeAPI/pull/1479)

### DIFF
--- a/src/main/java/org/spongepowered/mod/service/world/SpongeChunkTicketManager.java
+++ b/src/main/java/org/spongepowered/mod/service/world/SpongeChunkTicketManager.java
@@ -150,11 +150,13 @@ public class SpongeChunkTicketManager implements ChunkTicketManager {
         private PluginContainer plugin;
         private String pluginId;
         private ImmutableSet<Vector3i> chunkList;
+        private World world;
 
         private SpongeLoadingTicket(Ticket ticket) {
             this.forgeTicket = ticket;
             this.plugin = SpongeImpl.getGame().getPluginManager().getPlugin(ticket.getModId()).get();
             this.pluginId = this.plugin.getId();
+            this.world = (World) ticket.world;
         }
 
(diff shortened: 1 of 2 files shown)


adpking:
_Note:_ This will not build on Travis without the API change I made in another pull request.

liach:
@cossacksman You need to send a pull to common also.

adpking:
@liach Instead of making a PR for the API separately, I should just forget that PR and `git add SpongeCommon` to this commit with the changes, shouldn't I?

src/main/java/org/spongepowered/mod/service/world/SpongeChunkTicketManager.java
```diff
@@ -178,6 +180,11 @@ public int getMaxNumChunks() {
         }
 
         @Override
+        public World getWorld() {
+            return this.world;
+        }
```
liach, line 185: Travis says this override is invalid.

Mumfrey: Only because he forgot to commit the submodule update.

adpking: I've gone through the process again, updating the SpongeCommon and SpongeAPI submodules to point to my forked code; if you want you can disregard this pull and I can make another. I did this last night at 4am before a work day, so it was a little rushed.

Deamon5550:
Apparently moved to #1252 ...

PR #1252, opened 2017-02-06
Get the world from a LoadingTicket

adpking:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1480) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1190) | [SpongeForge](https://github.com/SpongePowered/SpongeForge/pull/1252)


Mohron:
@bloodmc would you support this being added since retrieving the world from a Sponge LoadingTicket is the only missing piece preventing implementation of a chunk loading plugin with only SpongeAPI?

adpking:
@Mohron Either that, or the Data API

PR #1593, opened 2017-06-17
Implement Recipe API

Faithcaio:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1582)|[SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1395)|*SpongeForge*

Register recipes in forge.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/common/MixinSpongeImplHooks.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/common/MixinSpongeImplHooks.java
@@ -34,6 +34,7 @@ import net.minecraft.entity.EnumCreatureType;
 import net.minecraft.entity.IEntityLivingData;
 import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.entity.player.EntityPlayerMP;
+import net.minecraft.item.crafting.IRecipe;
 import net.minecraft.launchwrapper.Launch;
 import net.minecraft.tileentity.TileEntity;
 import net.minecraft.util.EnumFacing;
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/mod/mixin/core/common/MixinSpongeImplHooks.java
```diff
@@ -286,4 +293,23 @@ public static void onTileEntityError(TileEntity tileEntity, CrashReport crashRep
     public static void blockExploded(Block block, World world, BlockPos blockpos, Explosion explosion) {
         block.onBlockExploded(world, blockpos, explosion);
     }
+
+    // Crafting
+
+    @Overwrite
```
stephan-gh, line 299: Please add a Javadoc comment with `@author` and `@reason` for both overwrites (Mixin will print compile warnings if they're missing)

stephan-gh (changes requested):

src/main/java/org/spongepowered/mod/mixin/core/common/MixinSpongeImplHooks.java
Faithcaio: I can do that but fyi no overwrite in this class has the comments

stephan-gh: Oh, this is `SpongeImplHooks`... Nevermind, the purpose of all the overwrites in this class is clear, so I don't think we need them in this case.

Faithcaio:
https://github.com/SpongePowered/SpongeCommon/commit/2a7a3664f2acbb92c1a5750ad288b1e242986e58

Issue #1049, opened 2016-11-19
Forge permissions API


Xfel:
Now that MinecraftForge provides a permissions API (https://github.com/MinecraftForge/MinecraftForge/pull/3155), SpongeForge should integrate with it, allowing us to query permissions from the mod's side.

ST-DDT:
I'm not sure what you are requesting. 
- Do you want SpongeForge to provide an implementation of the ForgePermission API 
- or do you want SpongeForge to default to Forge's Permission API 
- or both?

What happens if both Forge and Sponge provide a permission plugin?

There are also some fundamental differences between both concepts.
AFAICT there is no "OP" flag in Sponge, but it is used in Forge.

Xfel:
The basic permission check should be interoperable. As far as I can see, it would be pretty difficult to implement the sponge permissions api on top of the forge one, as the sponge api provides a lot more features. 

The other way round seems possible though. The only problem here is forge's DefaultPermissionLevel using the OP concept, which needs to be mapped to those default roles in some way.

Xfel:
I don't think this is related to commands specifically, the API is meant to be used by mods in general.

yuuka-miya:
Hi @ST-DDT and @zml2008 - 

I know this is fairly late (dropped out of the modding scene for quite a while), but I was one of those who helped @LatvianModder design the Forge permissions API. Just came across this issue, so I thought I'd share my thoughts on the matter.

My ideas for SpongeForge was that it provide a single **non-replaceable** implementation of the Forge permissions API, that delegates to Sponge's own permission service. This has the side effect of making SpongeForge automatically incompatible with other Forge-based permission providers, however I personally think that access to the more powerful Sponge permission system is worth the tradeoff in this case.

That way, both Forge mods and Sponge plugins get to be first-class citizens in a multiplayer server world, since now not only can Forge mods be permission aware, they can also take advantage of more powerful Sponge-based permission systems.

me4502:
I disagree with having it non overridable. If that were to be done, it'd make sense to have an option to disable it and a warning in the console if another forge permission mod is found.

maxanier:
+1 would love to register and check Sponge compatible permissions from within my Forge mod.

LatvianModder:
@me4502 the Forge itself already displays warning when a permission handler is replaced with another

Zidane:
@kashike 

Status?

kashike:
I'll get this done tonight most likely.

[...]

PR #1660, opened 2017-07-18
Add Forge permission handler

kashike:
[SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1461) | **SpongeForge**
#1049

### DIFF
--- a/src/main/java/org/spongepowered/mod/service/permission/SpongePermissionHandler.java
+++ b/src/main/java/org/spongepowered/mod/service/permission/SpongePermissionHandler.java
@@ -1,0 +1,102 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
+ * in the Software without restriction, including without limitation the rights
+ * to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
+ * copies of the Software, and to permit persons to whom the Software is
+ * furnished to do so, subject to the following conditions:
+ *
+ * The above copyright notice and this permission notice shall be included in
+ * all copies or substantial portions of the Software.
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.mod.service.permission;
+
+import com.mojang.authlib.GameProfile;
+import net.minecraftforge.server.permission.DefaultPermissionLevel;
+import net.minecraftforge.server.permission.IPermissionHandler;
+import net.minecraftforge.server.permission.PermissionAPI;
+import net.minecraftforge.server.permission.context.IContext;
+import org.spongepowered.api.Sponge;
+import org.spongepowered.api.entity.living.player.Player;
+import org.spongepowered.api.service.ProviderRegistration;
+import org.spongepowered.api.service.permission.PermissionDescription;
+import org.spongepowered.api.service.permission.PermissionService;
+import org.spongepowered.api.service.user.UserStorageService;
+import org.spongepowered.api.text.Text;
+import org.spongepowered.common.SpongeImpl;
+import org.spongepowered.common.registry.RegistryHelper;
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.HashSet;
+import java.util.Set;
+
+import javax.annotation.Nullable;
+
+// https://github.com/SpongePowered/SpongeForge/issues/1049#issuecomment-269775514
+public enum SpongePermissionHandler implements IPermissionHandler {
+    INSTANCE;
+
+    private final Set<String> nodes = new HashSet<>();
+
+    public void adopt() {
+        PermissionAPI.setPermissionHandler(this);
+    }
+
+    public void forceAdoption() {
+        SpongeImpl.getLogger().debug("Forcing adoption of the Sponge permission handler - previous handler was '{}'", PermissionAPI.getPermissionHandler().getClass().getName());
+        RegistryHelper.setFinalStatic(PermissionAPI.class, "permissionHandler", this);
+    }
+
+    @Override
+    public void registerNode(final String node, final DefaultPermissionLevel level, final String desc) {
+        this.nodes.add(node);
+        // TODO: do.... what with level and desc?
+    }
+
+    @Override
+    public Collection<String> getRegisteredNodes() {
+        return Collections.unmodifiableSet(this.nodes);
+    }
+
+    @Override
+    public boolean hasPermission(final GameProfile profile, final String node, @Nullable final IContext context) {
+        @Nullable final Player player = Sponge.getServer().getPlayer(profile.getId()).orElse(null);
+        if (player != null) {
+            // TODO contexts?
+            return player.hasPermission(node);
+        }
+
+        // TODO contexts?
+        return Sponge.getServiceManager().getRegistration(UserStorageService.class)
+                .map(ProviderRegistration::getProvider)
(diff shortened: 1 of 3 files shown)


kashike:
@luacs1998: Could I get your input here? You're familiar with the Forge permission API.

src/main/java/org/spongepowered/mod/command/SpongePermissionHandler.java
```diff
@@ -0,0 +1,89 @@
...
+        try {
+            PermissionAPI.setPermissionHandler(this);
+        } catch (IllegalStateException ignored) {
+            RegistryHelper.setFinalStatic(PermissionAPI.class, "permissionHandler", this);
+        }
+    }
+
+    @Override
+    public void registerNode(final String node, final DefaultPermissionLevel level, final String desc) {
+        // TODO
```
yuuka-miya: Does Sponge have the concept of registering permissions?

Still trying to acquaint myself with the Sponge PermissionService, so...

yuuka-miya:
Looks okay to me, but if you want a second opinion - @LatvianModder?

src/main/java/org/spongepowered/mod/command/SpongePermissionHandler.java
kashike: We have the concept of [creating descriptions](https://github.com/SpongePowered/SpongeAPI/blob/bleeding/src/main/java/org/spongepowered/api/service/permission/PermissionService.java#L94-L102) which requires the mod/plugin, the permission node, and a description.

src/main/java/org/spongepowered/mod/command/SpongePermissionHandler.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public boolean hasPermission(final GameProfile profile, final String node, @Nullable final IContext context) {
+        @Nullable final Player player = Sponge.getServer().getPlayer(profile.getId()).orElse(null);
+        if (player == null) {
+            return false;
+        }
+
+        // TODO contexts?
```
kashike: How should we handle this? Sponge has a `Context` concept as well, not sure if we want to convert `IContext` data to a `Context`?

src/main/java/org/spongepowered/mod/command/SpongePermissionHandler.java
```diff
@@ -0,0 +1,89 @@
...
+    }
+
+    @Override
+    public void registerNode(final String node, final DefaultPermissionLevel level, final String desc) {
+        // TODO
+    }
+
+    @Override
+    public Collection<String> getRegisteredNodes() {
+        return Collections.emptySet(); // TODO
```
kashike: Should we keep track of this, just for Forge mods wishing to get this?

src/main/java/org/spongepowered/mod/command/SpongePermissionHandler.java
ryantheleach: To add on to this, descriptions can designate generic like permissions with placeholders for variable permissions. e.g. nucleus.teleport.world.<dimension> to have per dimension permissions.

LatvianModder:
I guess this works. I wouldnt use that many lambdas in getNodeDescription, but that isn't used frequently anyway, so I guess its fine. Im ok, as long as it properly handles registered permissions. Only not sure why is that try{}catch{} required, just register your handler in PREINIT, and it will be fine, because mods are expected to register nodes in INIT+

src/main/java/org/spongepowered/mod/command/SpongePermissionHandler.java
```diff
@@ -0,0 +1,89 @@
...
+    // https://github.com/SpongePowered/SpongeForge/issues/1049#issuecomment-269775514
+    public void adopt() {
+        if (PermissionAPI.getPermissionHandler() == this) {
+            return;
+        }
+
+        try {
+            PermissionAPI.setPermissionHandler(this);
+        } catch (IllegalStateException ignored) {
+            RegistryHelper.setFinalStatic(PermissionAPI.class, "permissionHandler", this);
```
Zidane: This is bad, if setting the permission handler fails then log it to the console as that likely means someone else is doing so. Let the server admin figure out how to proceed but don't blindly blow away another mod whose purpose is to do permissions.

LatvianModder: No, I actually agree with Sponge being the strongest permission handler (Config option for this would be nice tho), but I dont agree with aggressively changing the private field. You only should set permission handler in PREINIT state, as I said earlier. Hopefully this is changed

kashike: @luacs1998 also suggested that we be the strongest. There's the possibility that another mod will replace our handler, so my thought was to do it after pre-init. I can add a config option to *disable* our replacing of other permission handlers if one is registered before Sponge.

src/main/java/org/spongepowered/mod/service/permission/SpongePermissionHandler.java
```diff
@@ -0,0 +1,87 @@
...
+ *
+ * THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.mod.service.permission;
```
dualspiral: I put this on a commit that has been blown away rather than the PR, but I'll put it in again - permissions have nothing to do with commands, why is this in a commands package?

src/main/java/org/spongepowered/mod/service/permission/SpongePermissionHandler.java
```diff
@@ -0,0 +1,87 @@
...
+    }
+
+    @Override
+    public Collection<String> getRegisteredNodes() {
+        return Collections.emptySet(); // TODO
+    }
+
+    @Override
+    public boolean hasPermission(final GameProfile profile, final String node, @Nullable final IContext context) {
+        @Nullable final Player player = Sponge.getServer().getPlayer(profile.getId()).orElse(null);
```
Xfel: This will only work for online players. Is that intended?

Zidane: @kashike ?

phit:
@kashike status of this?

BrainStone:
Yes, what’s the status on this? We still need a plugin for this.

BrainStone:
Bump...

tonnic:
please add this....we need this implemented

maxanier:
What exactly has to be tested?
I could try to verify the general functionality by adding some Forge permission checks to a mod of mine. Then build this branch, setup a server with it, my mod and LuckPerms and check if LuckPerms can access the mod's permissions?

gabizou:
> I could try to verify the general functionality by adding some Forge permission checks to a mod of mine. Then build this branch, setup a server with it, my mod and LuckPerms and check if LuckPerms can access the mod's permissions?

Correct

maxanier:
So, I did that and (after I found out I have to enable this in the Sponge configuration) it works fine. I can give/deny the permissions registered by my mod using LuckPerms and they are correctly identified.

Only issue is that, the default permission level passed to the registerNode method is ignored and instead the default permission level is undefined (which results in deny).
Most, if not all, permissions introduced by my mod should be available to all players. So one would have to manually grant "vampirism.*" to every player/group. Before they are actually able to play the mod with this enabled.

However, since this has to be manually enabled anyway, I guess that is ok.
I am very happy there is finally a way to integrate mods with permission plugins.

phit:
you could already do so https://docs.spongepowered.org/stable/en/plugin/permissions.html#for-forge-mods you don't need this PR for that

LatvianModder:
If I had known about this way before the Forge PermissionAPI PR was merged, I would have simply just added a hook in checkPermission() method for forge mods, which would make things 10000x easier. That's why I suggest to try this approach in 1.13/1.14+ and basically just not care about it in 1.12. Mods would most likely not care about it this late anyway, but if new PR was made for 1.13, maybe then things can change

BrainStone:
The Forge Permission API is pretty well used already on 1.12.2. So it would make sense to already implement it in API 7, so people don't have to use an unreleased version of an abandonned plugin that bridges the gap.  
Keep in mind that 1.12.2 will be a long term version for modded MC.

BrainStone:
@gabizou what's the holdup? This PR is a strict improvement of the project, has been confirmed to work and been reviewed by several people.

Zidane:
@phit has objections to merging this and can elaborate why.

BrainStone:
Sure. Let's keep it rotting until it can't be merged anymore.

pie-flavor:
@BrainStone oy. No need to be rude.

ImMorpheus (approved):
Tested with FTB Utilities.
Tried to change `ftbutilities.chat.nickname.set` with Luckperms, everything is working as intended.

PR #1723, opened 2017-08-14
Method Tracking System - ItemHandler and Fluid tracking

Cybermaxke:
[Common](https://github.com/SpongePowered/SpongeCommon/pull/1483) | **Forge**

This PR uses the newly added Method Tracking System to track changes in `IItemHandler`, `IFluidTank`, `IFluidHandler` and `IFluidBlock`. Only the framework to modify these methods is added, currently are no extra actions added to the redirected methods.


gabizou:
Looking at this code, I stumbled on it and asked myself "What are we doing here?" Did we have intentions to track causes and throw events for these sort of interfaces? Or is it just provided for GriefPrevention's uses?

Pinging @SpongePowered/developers for some input.

JBYoshi:
I think that was the intent.

PR #1823, opened 2017-10-29
Expand and implement the Instrument/Noteblock API.

Cybermaxke:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1608) | [Common](https://github.com/SpongePowered/SpongeCommon/pull/1437) | **Forge**

Prevent forge from blocking the `PLING` instrument.

### DIFF
--- a/src/main/java/org/spongepowered/mod/event/NoteBlockPlayEvent.java
+++ b/src/main/java/org/spongepowered/mod/event/NoteBlockPlayEvent.java
@@ -1,0 +1,46 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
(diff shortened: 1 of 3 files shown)


ryantheleach:
@Cybermaxke Does forge block this deliberately? This seems like something better fixed upstream.

Cybermaxke:
@ryantheleach I think it's better to remove the ping sound, it's not officially supported and the instruments will also be an enum in 1.13. The instrument will become a part of the block state of a note block, (and tile entity gets removed), the overridden instrument will also be broken, since you can't store the instrument anymore in the tile entity. You can still change the instrument in the block state, but that will be updated when placing a new block under it.

Cybermaxke:
Closing, ping instrument will be broken in 1.13 so there is no need to add any more breaking changes.

PR #1839, opened 2017-11-07
Refactor Github issue template

phit:
Tried to simplify it a bit and reworded a few things. I also got rid of the OS checkbox, it almost never adds any value and gets handled as tasks by Github, which leads to these misleading UI changes.

image

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -1,24 +1,21 @@
-<!-- Please be as accurate as possible, and include as much information as possible.
-         Add or remove as needed to keep things informative but also relevant. -->
-
-**I am currently running...**
+**I am currently running**
 <!-- If you don't use the latest version, please tell us why. -->
 - SpongeForge version: 
 - Forge version: 
 - Java version: 
 - Operating System:
-<!-- Replace the empty space (" ") in with an "x" within the brackets. -->
- - [ ] Linux (Ubuntu, Debian, Fedora, etc)
- - [ ] macOS
- - [ ] Windows
-<!-- You can get a list of plugins/mods by running the `/sponge plugins` command on your server -->
-- Plugins/Mods: 
 
-**What is the expected result?**
+<!-- Please include ALL mods/plugins you had installed when your issue happened, you can get a list of 
+        your mods and plugins by running "/sponge plugins" and/or "/sponge mods" -->
+- Plugins/Mods:
 
-**What is the current result?**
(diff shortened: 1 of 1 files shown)


Zidane:
@phit 

Couple of changes.

* I've found that dealing with downloading a text file and reading it really makes for an annoying investigation process. Honestly, they need to put it in gist, always. They never expire and commenting here means they already have a gist. It also makes it much easier to read on mobile.
* I think the part with plugins/mods needs to be elaborated on. While I see where you are going with the language used for getting a plugins/mods list, I'm afraid that people will just put what plugins/mods they THINK is doing it without actually running their server/client WITH ONLY those plugin/mods they are giving us. Thoughts?

phit:
* Fine with me, I'll change it to only accept gist.github.com
* I'm torn on this, because right now people just copy&paste their modlist, which is super useless and you have to figure out what mod is actually needed to reproduce. I think only asking for the mods they think are responsible is fine. If we need a more detailed modlist you can always check the logs they have provided, which also has the benefit of accurate version info.

Zidane:
@phit I can compromise and state that if we are given a log that shows that they are running additional mods/plugins the issue will be auto-closed as that means they are lying about their issue. The template needs to state that you have to give us a log with just these mods/plugins to prove the issue does happen or show what is going on with that issue in the logs with what is provided.

jamierocks:
> I can compromise and state that if we are given a log that shows that they are running additional mods/plugins the issue will be auto-closed as that means they are lying about their issue.

This would be fine if this was an issue tracker exclusively for developers, but its not - and not everyone knows how to read stacktraces. So, I would certainly have a full list of installed mods. I agree with @phit's point regarding useless information - but server administrators might not know which mods causes more issues than others, and it could even differ by experience. 

IMO, there is no good approach here, other than to request a full mod list - it would certainly be a bad thing to close an issue based on someone's opinion being wrong, and conflicting with a stack trace they cannot read.

Zidane:
@jamierocks Except this has nothing to do with stacktraces. If someone says "Oh Nucleus and Sponge only installed causes a crash! Here is my full log!" and we see that it has 2 other plugins/mods...the issue will be closed and is NOT up for discussion.

I'm not wasting man hours by having one of my guys go on a witch hunt debugging an admin's server installation for them.

phit:
bump, anything else that needs changing?

Zidane:
@phit can you rebase into one commit?

phit:
@Zidane done!

kashike (approved):

PR #1858, opened 2017-11-16
SC1523: Stop Leaf Decay from leaking.

ryantheleach:
This adds a cause frame to leaf decay to stop it from leaking. see 

https://github.com/SpongePowered/SpongeCommon/issues/1523
https://github.com/SpongePowered/SpongeCommon/issues/1530
etc.

Should this be making a cause frame at this level? Or higher up the callstack?

Additionally, is this appropriate for SF? I was expecting this to be in common.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/core/block/MixinBlockLeaves.java
+++ b/src/main/java/org/spongepowered/mod/mixin/core/block/MixinBlockLeaves.java
@@ -30,6 +30,7 @@ import net.minecraft.block.state.IBlockState;
 import net.minecraft.util.math.BlockPos;
 import org.spongepowered.api.Sponge;
 import org.spongepowered.api.block.BlockState;
+import org.spongepowered.api.event.CauseStackManager;
 import org.spongepowered.api.event.cause.EventContextKeys;
 import org.spongepowered.api.util.annotation.NonnullByDefault;
 import org.spongepowered.api.world.LocatableBlock;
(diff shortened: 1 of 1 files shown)


Deamon5550 (approved):

src/main/java/org/spongepowered/mod/mixin/core/block/MixinBlockLeaves.java
```diff
@@ -53,26 +54,28 @@
...
+                if (isBlockAlready && !isWorldGen) {
+                    final LocatableBlock locatable = LocatableBlock.builder()
+                            .location(new Location<World>((World) worldIn, pos.getX(), pos.getY(), pos.getZ()))
+                            .state((BlockState) state)
+                            .build();
+                    BlockPhase.State.BLOCK_DECAY.createPhaseContext()
+                            .source(locatable)
+                            .buildAndSwitch();
+                }
+                block.beginLeavesDecay(state, worldIn, pos);
```
gabizou, line 75: This should be wrapped in a try with resources for the phase switch.

gabizou (changes requested):
minor change and should be good to go. May want to update SpongeForge specific mixing for leaf decay as well.

ryantheleach:
This is the SF specific mixin, Do you know for what purpose they were separated?

src/main/java/org/spongepowered/mod/mixin/core/block/MixinBlockLeaves.java
ryantheleach: Thanks for the advice I'm new to the phase stuff, but isn't the problem is the conditionals before and afterwards. 

All I've done is wrap everything in the if in a TryResource ContextFrame block.

I assume the !isWorldGen is to prevent spam during world generation as well as keeping the phase rules for world generation.

The isBlockAlready I don't quite understand.

gabizou:
> Do you know for what purpose they were separated?

Because the method they are being mixed into only exists in Forge ;)

PR #1896, opened 2017-12-13
Fix incorrect unit test directory in Gradle project and failing test

ghost:
Replaces #964 - My original fork was removed, since it was last updated in October 2016. Please see the discussion from over there.

-----

#### As discussed in IRC:

Gradle/Maven expects unit tests to be found at `src/test/java`, however SpongeForge has been using `src/main/test/java`. This is incorrect.

---

Appended class with `@PowerMockIgnore` annotation to prevent PowerMockito from mocking classes we don't want to mock.

Also adds another `EventManager#registerListeners()` call in `testSpawn()`, as this **_should_** be expected behavior (we don't want to fire an event if nothing is listening for it - wasted processing time). As discussed on the previous MR, this behavior might not have the correct intentions. Would appreciate it if someone could clarify this is what we want, I'll happily edit.


kashike (approved):

src/test/java/org/spongepowered/mod/event/SpongeModEventManagerTest.java
```diff
@@ -25,10 +46,10 @@
 
 import java.lang.reflect.Field;
 import java.lang.reflect.Modifier;
-import java.net.URLClassLoader;
 import java.util.Optional;
 
 @RunWith(PowerMockRunner.class)
```
JBYoshi, line 51: This doesn't compile. There's no Powermock dependency in Gradle.

JBYoshi (changes requested):
In addition to what I marked below, the `SpongeEventManager` constructor now needs a `Logger` object (GitHub doesn't let me comment on that line since it's not part of the main diff).

src/test/java/org/spongepowered/mod/event/SpongeModEventManagerTest.java
```diff
@@ -75,6 +96,7 @@ public void testSpawn() throws ClassNotFoundException {
 
         eventBus.register(forgeListener);
 
+        this.eventManager.registerListeners(this.plugin, listener);
```
JBYoshi, line 99: I don't know if this is supposed to be there. Sponge relays some of its own events (such as `SpawnEntityEvent`) back to Forge. While there are no Sponge listeners, the event still needs to be fired for the Forge listeners. This messes with the test by adding a Sponge listener when the test should only have Forge listeners at this point.

Zidane:
Closing due to no response.

PR #1913, opened 2017-12-25
Advancements implementation

Cybermaxke:
[API](https://github.com/SpongePowered/SpongeAPI/pull/1554) | [Common ](https://github.com/SpongePowered/SpongeCommon/pull/1645)| **Forge** | [Vanilla](https://github.com/SpongePowered/SpongeVanilla/pull/363)


src/main/java/org/spongepowered/mod/mixin/core/fml/common/crafting/MixinCraftingHelper.java
```diff
@@ -0,0 +1,61 @@
...
+    private static File onFindFiles(ModContainer mod) {
+        final File source = mod.getSource();
+        if (!source.isDirectory() || new File(source, "assets").exists()) {
+            return source;
+        }
+        // TODO: Clean way to handle this? Fix this in forge?
+        // In a development environment are the resources and classes separated
+        // into two source directories, at least in IntelliJ.
+        // This is a temporary fix to make sure that advancement/recipe files
+        // are loaded in the IDE.
```
Cybermaxke: I would love some input on this ugly fix, no idea how to work around it.

stephan-gh: Is this for SpongeForge itself or for Sponge plugins loaded from the classpath?

Cybermaxke: Sponge plugins loaded from the classpath.

stephan-gh: Generally, we set `inheritOutputDirs = true` for all IntelliJ IDEA modules, which makes IntelliJ IDEA write classes and resources into one shared output directory. This is applied automatically for SpongeAPI and the Sponge implementation modules and all plugins using SpongeGradle.

However, I noticed that it wasn't applied to `testplugins` which might be the reason for the problem if that's what you were testing with. I fixed that in https://github.com/SpongePowered/SpongeAPI/commit/25890783ac54848846bfe21e1542dfbf5b8e38e2, can you check if your workaround is still necessary after this commit?

If you are using a separate plugin for testing, make sure to apply the SpongeGradle plugin.

Cybermaxke: That worked, thanks! :)

PR #1929, opened 2018-01-02
Fix event generics

Cybermaxke:
[Common](https://github.com/SpongePowered/SpongeCommon/pull/1681) | **Forge**

Also fixes a issue that caused the cause stack manager to be used on the client when it's not supposed to.

### DIFF
--- a/src/main/java/org/spongepowered/mod/event/SpongeModEventManager.java
+++ b/src/main/java/org/spongepowered/mod/event/SpongeModEventManager.java
@@ -351,15 +351,7 @@ public class SpongeModEventManager extends SpongeEventManager {
     }
 
     @Override
-    public boolean post(Event event) {
-        return this.post(event, false);
-    }
-
-    @Override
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/mod/event/SpongeModEventManager.java
```diff
@@ -351,15 +351,7 @@ protected boolean post(Event event, List<RegisteredListener<?>> listeners, boole
     }
 
     @Override
-    public boolean post(Event event) {
-        return this.post(event, false);
-    }
-
-    @Override
-    public boolean post(Event spongeEvent, boolean allowClientThread) {
-        if (!allowClientThread & Sponge.getGame().getPlatform().getExecutionType().isClient()) {
```
gabizou: For all sponge events, this is honestly preferred that the client thread check is done. Because if we are allowing client thread events for events that are normally not client thread events, there's going to be issues with how the events are thrown and handled, with the exception of pre-server starting.

PR #1995, opened 2018-02-10
Tick the scheduler when in client mode as well.

Dockter:
Signed-off-by: Dockter <dockter@almuramc.com>

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeMod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeMod.java
@@ -276,8 +276,10 @@ public class SpongeMod extends MetaModContainer {
     @SideOnly(Side.CLIENT)
     @SubscribeEvent
     public void onTick(TickEvent.ClientTickEvent event) {
-        if (event.phase == TickEvent.Phase.START) {
-            this.scheduler.tickSyncScheduler();
+        if (!net.minecraft.client.Minecraft.getMinecraft().isIntegratedServerRunning()) {
+            if (event.phase == TickEvent.Phase.START) {
+                this.scheduler.tickSyncScheduler();
+            }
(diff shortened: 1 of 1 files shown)


kashike:
~qa

Dockter:
Added check so the client isn't double ticking the queue.

src/main/java/org/spongepowered/mod/SpongeMod.java
```diff
@@ -276,8 +276,10 @@ public void onTick(TickEvent.ServerTickEvent event) {
     @SideOnly(Side.CLIENT)
     @SubscribeEvent
     public void onTick(TickEvent.ClientTickEvent event) {
-        if (event.phase == TickEvent.Phase.START) {
-            this.scheduler.tickSyncScheduler();
+        if (!net.minecraft.client.Minecraft.getMinecraft().isIntegratedServerRunning()) {
```
Zidane: Fully qualified import is un-needed as import lines are just for our benefit. When the code is compiled, everything is made fully qualified.

Dockter: TIL, ty.

Zidane:
Manually merged.

simon816:
I'm not sure about this... `ClientTickEvent` occurs on the client thread, running the scheduler on any thread other than the server thread will cause issues.

I'm not sure why you want to be running server tasks outside of the server lifecycle. This seems like something you'd want from a client API, operating on a separate client-side scheduler

Zidane:
@simon816 Hasn't caused any issues in the two weeks this has been live. I'm well aware that a Client scheduler should be added but it isn't a priority to do that refactor currently.

Issue #1324, opened 2017-03-04
Items despawning that should not


ghost:
I don't know if this is the right place for this, but I believe it would be SF that's causing items that shouldn't despawn (Like Botania's Terra Shatterer) to do so. How might I track this down? Any help with this would be greatly appreciated.

Twisted-Code:
yeah same thing is happening on my server with Applied Energistics' crystal seeds, though I haven't been able to reproduce it with ONLY AE and SF installed, so I think it might be a bad interaction between SF and something else. What mods do you have installed? I'll let you know if any overlap with my pack and we can go from there

phit:
dupe of #1201

Twisted-Code:
okay whatever the cause of my specific issue is, it doesn't happen on single player. Still could be a bad interaction between SF and a mod, but what mod?

ryantheleach:
@macks2008 Thanks for this update, reports like this are vital information when developers attempt to debug.

phit:
still happens with spongeforge-1.12.2-2503-7.0.0-BETA-2716

Forge's EntityItem#setNoDespawn() is being ignored, see https://github.com/SpongePowered/SpongeForge/issues/1201#issuecomment-325281084

LemADEC:
Do we have any idea how to fix that issue?
Any whitelisting configuration we could have in the mean time?

phit:
someone would have to implement checking for the forge ´lifespan´ tag, I looked into it before but no idea how to best integrate this as the item despawning code is in SC, but the ´lifespan´ tag obviously isnt a thing thing in SV

LemADEC:
Could we have a whitelist or a way to disable the feature for now?

phit:
there is no feature or anything in place, infinite lifespan set by forge is simply never checked
adding a whitelist would be more effort than fixing the actual issue

LemADEC:
By feature, I mean the sponge part that wants to override the despawning mechanism

phit:
disable the entity activation module

ProsperCraft:
AE2 nether quartz seeds which are never supposed to despawn are still despawning in spongeforge-1.12.2-2611-7.1.0-BETA-2934.jar

ryantheleach:
The linked PR was merged into bleeding. It will end up in a Sponge release eventually, for now disable the entity activation module.

ProsperCraft:
AE2 seeds are still despawning with entity activation mode on in 2996

PR #2000, opened 2018-02-10
Fix lifespan being ignored when EntityItem Activation is enabled

phit:
Forge adds a lifespan NBT tag allowing mods to set a longer despawn
delay than possible in Vanilla, Sponge previously ignored this when
ticking inactive items. This adds a check if items have the default
lifespan, fall back to the Sponge config for maximum age, otherwise use
the value set by mods.

Fixes https://github.com/SpongePowered/SpongeForge/issues/1324


phit:
~qa

PR #2060, opened 2018-03-23
Fix build.gradle using reference to rootProject.

### LINKED ISSUE #2049: SpongeForge shouldn't be referenced as rootProject
**I am currently running**

- SpongeForge version: all
- Forge version: n/a
- Java version: n/a
- Operating System: n/a

- Plugins/Mods: n/a

**Issue Description**
SpongeForge should not assume that it is the root gradle project, because it doesn't have to be.
It makes that assumption here: https://github.com/SpongePowered/SpongeForge/blob/bleeding/build.gradle#L89

This means if you try include SpongeForge in a wrapper gradle module, this line will fail, because `rootProject` now points somewhere else.

Instead of `rootProject`, it should just use `this`, so it always points to SpongeForge

gravityfox:
Replace `rootProject` reference with `this` to make SpongeForge behave when it isn't the root project.
closes #2049

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -86,7 +86,7 @@ artifacts {
 
 allprojects {
     // Re-obfuscate only after creating the shadowDevJar
-    tasks.findByName('reobfJar')?.mustRunAfter rootProject.tasks.shadowDevJar
+    tasks.findByName('reobfJar')?.mustRunAfter this.tasks.shadowDevJar
 }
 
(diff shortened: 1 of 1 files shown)


gravityfox:
Actually, this would be convenient if it could be backported to earlier branches (to make them easier to work with), cause it's just a single line of gradle that hasn't been touched in awhile, but i'll leave it on bleeding for now.

PR #2110, opened 2018-04-20
Refactor Entity Drops for Forge Mod Compatibility

gabizou:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1808) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1882) | **SpongeForge**

This brings maximum compatibility of Sponge's tracking system with Forge's capture drops for entities and this time, properly handles custom entities from item drops.

This is still a work in progress.


gabizou:
@Thutmose I've verified now that the custom item drops do work, regardless whether they're dropped from players or from deaths (I'm guessing this is the intended function, yes?)

Thutmose:
ya, the intended behavior is whenever an itemstack of that item is made into an EntityItem, instead of spawning the EntityItem, it spawns the custom entity instead, so in the case of the test mod, a chicken.

This should happen from being dropped intentionally, via death, via droppers, etc, at least that is what happens from the "vanilla" behavior with just forge

gabizou:
Awesome, I've inlined it into sponge's capture system such that any entityitem spawned will be checked for having a custom item (of course this is done without needing to have forge listen for an event to perform the action). 

I'm almost done cleaning up the mess and verifying I'm leaving the bit of codebase in better condition (maintainability) than I found it.

Issue #1991, opened 2018-02-08
Failed to save configuration


ghost:
- SpongeForge version: spongeforge-1.12.2-2586-7.1.0-BETA-2887.jar
- Forge version: 14.23.1.2586
- Java version: build 1.8.0_151-b12
- Operating System: 64bit windows 10 home

Plugins: Currently none
Mods: Minecraft, Minecraft Coder Pack, Forge Mod Loader, Minecraft Forge, SpongeAPI, SpongeForge, AE2 Stuff, Actually Additions, Advanced Generators, Ambience, Applied Energistics 2, Aroma1997Core, AromaBackup, AromaBackuprecovery, AutoRegLib, BD Lib, Baubles, Better Builder's Wands, BetterBedrockGen, Block Armor, Botania, ChickenChunks, Chisel, Chunk Pregenerator, CoFH Core, CoFH World, CodeChicken Lib, Compact Solar Arrays, ComputerCraft, CraftTweaker JEI Support, CraftTweaker2, Cucumber, Cyberware, DeepResonance, ElecCore, Extended Crafting, External Tweaker, Extra Cells 2, Extra Utilities 2, Extreme Reactors, Gravestone Mod, ICBM-Classic, Industrial Foregoing, IndustrialCraft 2, InstantUnify, Iron Chest, IvToolkit, Loot Bags, LootTweaker, MTLib, Magical Psi, Mantle, Mekanism, MekanismGenerators, MekanismTools, MmmMmmMmmMmm, Moar Tinkers, Mod Tweaker, Modern Warfare, OMLib, Open Modular Turrets, OpenBlocks, OpenComputers, OpenComputers (Core), OpenModsLib, OpenModsLib Core, OreExcavation, PlusTiC, ProjectE, Psi, Quick Leaf Decay, Random Things, Redstone Arsenal, Redstone Flux, Refined Storage, SanAndreasPs Mod Library, SanAndreasPs Player Model, Secret Rooms 5, Shadowfacts' Forgelin, SimpleInventoryStarter, TATW, Tesla Core Lib, Tesla Core Lib Registries, The Lost Cities, Thermal Cultivation, Thermal Dynamics, Thermal Expansion, Thermal Foundation, Tinker I/O, Tinkers Tool Leveling, Tinkers' Addons, Tinkers' Construct, UniDict, Voltz Engine, WanionLib, Zero CORE, mob_grinding_utils

[...]

PR #2162, opened 2018-05-19
Delay config saving to prevent Access Denied errors on Windows

dualspiral:
[SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1915) | **SpongeForge** | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/369) | [Original SC Issue](https://github.com/SpongePowered/SpongeCommon/issues/1906) | [Original Original SF Issue](https://github.com/SpongePowered/SpongeForge/issues/1991)

See the SpongeCommon PR for more information

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeMod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeMod.java
@@ -368,6 +368,7 @@ public class SpongeMod extends MetaModContainer {
     public void onPostInitialization(FMLPostInitializationEvent event) {
         try {
             SpongeImpl.getRegistry().postInit();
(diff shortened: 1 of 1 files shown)


Zidane:
@dualspiral 

If I'm splitting hairs here, you could only flush in init only if we're in client mode.

dualspiral:
@Zidane 

I'm inclined to agree with you, same with your SV comment. I'll update when I next have chance.

bloodmc:
Pulled
https://github.com/SpongePowered/SpongeForge/commit/3d9d99ed62592d567f6ec2e644c36a699c100f48

Thanks

PR #2164, opened 2018-05-19
Change absolute module path to relative module path.

gravityfox:
Fixes issues when including SF as submodule in gradle.


gravityfox:
actually hold this thought. I need to try something first

PR #2299, opened 2018-07-13
Add support for working around broken mods

Aaron1011:
**SpongeForge** | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/1991)

A large number of mods have improper network handlers. Instead of
interacting with the world from a task scheduled on the main thread,
they directly modify the world from a Netty handler. This is strictly
incorrect, and will cause issues even when Sponge is not installed.

However, due to the extra thread checks added by Sponge, mods which
*seem* to work correctly without Sponge can appear to break when it is
added. While the ultimate solution is always for the mod to only modify
the world from the main thread, mod developers don't always respond
quickly to these kinds of issues. The end result: while these kinds of
mods are broken even without Sponge, the extra checks done by Sponge
means that these mods may appear 'more broken' with Sponge installed.

As a temporary workaround to help server owners, this commit adds an
option to forcibly reschedule mod network handlers on the main thread.
To avoid messing with Netty directly, only mods using Forge's 'simple
channel' system are handled. While Forge technically provides other
channel options to mods, in practice virtually all mods use the 'simple
channel' system. Indeed, to date all issues opened on Sponge for these
misbehaving mods are caused by mods using the 'simple channel' system.

The implementation of this feature is minimally invasive - it substitues
a subclass of Forge's SimpleChannelHandlerWrapper when a mod registers
a 'simple channel' handler. However, to further minimize its impact,
this replacement is completely off by default - not even the mixin is
applied. Furthermore, this feature must be explicitly enabled for
specific mod ids via the config. Any mods not in the config will be
completely untouched.

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeCoremod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeCoremod.java
@@ -97,6 +97,7 @@ public class SpongeCoremod implements IFMLLoadingPlugin {
         }
 
         Mixins.addConfiguration("mixins.forge.core.json");
+        Mixins.addConfiguration("mixins.forge.brokenmods.json");
         Mixins.addConfiguration("mixins.forge.bungeecord.json");
         Mixins.addConfiguration("mixins.forge.entityactivation.json");
         Mixins.addConfiguration("mixins.forge.optimization.json");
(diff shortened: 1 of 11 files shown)


Aaron1011:
Manually merged in https://github.com/SpongePowered/SpongeForge/commit/dfa2a385f684ddb13e450f10a34c4121c1e44922

Issue #1838, opened 2017-11-07
Command blocks desync after switching dimension


phit:
- SpongeForge version: 1.10.2-2698 & 1.11.2-2484 & 1.12.2-2729
- Plugins/Mods: none

**What is the expected result?**
switch dimension through a nether portal and be able to place a commandblock and access it, without relogging or "opping yourself again"

**What is the current result?**
>when i teleport across two dimensions, after i do it i'm not able to place command blocks : i see them places but can't interact with thm and when i break it it make no particle like if it wasn't really here so i can't use it until i re-op myself even if i'm already ... 

> [logs + client side video of the bug.zip](https://github.com/SpongePowered/SpongeForge/files/1185300/logs.client.side.video.of.the.bug.zip)

originally reported by @Eldwen here https://github.com/SpongePowered/SpongeForge/issues/1684
just tested this again and confirmed this is still an issue

ryantheleach:
Closed in favor of SpongePowered/SpongeCommon#1718 (Reproducible on SV)

PR #2319, opened 2018-07-18
Add integration test support via McTester

Aaron1011:
This PR adds several integration tests to SpongeForge, using my project [McTester](https://github.com/Aaron1011/McTester).

**Overview**

McTester allows writing tests that run on an actual singleplayer client instance. These tests are able to control the client input, allowing them to perform actions like sending chat message, looking around, and left/right clicking. This allows writing tests that wouldn't otherwise be possible - for example:

* Igniting a creeper by right-clicking with a flint and steel, and checking that an event is fired with an appropriate Cause.
* Eating a food item on the client, and ensuring that the proper sequence of `UseItemStackEvent`s is fired.
* Ensuring that a GUI actually opens on the client when a player with the proper permissions right-clicks a command block.

These tests can interact with both the server and the client. Since tests don't run on the server thread, all access to the server must be done through `TestUtils.runOnMainThread`. Access to the client is done via the the special `Client` interface. More information is available on [the javadocs](https://jitpack.io/com/github/Aaron1011/McTester/0.5.4-SNAPSHOT/javadoc/)

One particularly powerful use of McTester is writing regression tests for issues that could otherwise only be tested manually. One example is https://github.com/SpongePowered/SpongeForge/issues/1838, where command blocks stopped working after switching dimensions. A proper test for this issue requires a running client, since the server has no idea if a player has a command block GUI open or not. With McTester, this issue can be tested in a completely automated fashion, ensuring that the issue doesn't reappear (as it did once already!).

**Tests**:

This PR adds the following tests. In addition to demonstrating the functionality of McTester, all of these tests are useful in their own right:

* ChatTest - tests that typing and sending a message on the client triggers an event on the server
* CreeperTest - tests that igniting and exploding a creeper fires a `ChangeBlockEvent` with a proper cause and context.
* InventoryTest - tests that giving the player an item through the Inventory API causes the item to show up on the client.
* UseItemStackTest - tests that eating an item fires the expected sequence of UseItemStackEvent sub-events
* BreakBlockUsedItem - tests that breaking an item on the client fires a ChangeBlockEvent with `EventConextKeys.USED_ITEM` in it
* CrossDimensionCommandBlock - tests that a player with the proper permissions can still open command blocks after changing dimensions.
* TntExplosionTest - tests that igniting TNT with flint and steel causes the proper explosion event to be fired
* UseItemStackFinishBug - tests that UseItemStackEvent.Finish is fired with the proper values

**Kotlin support**

One test in this PR is different from the others - [CreeperTestKotlin](https://github.com/Aaron1011/Sponge/blob/29a6a8d6326127109bd88ef8541b01a1edfd1d65/src/test/kotlin/org/spongepowered/mod/CreeperTestKotlin.kt). This demonstrates McTester's experimental support for writing tests using Kotlin coroutines.

Normally, McTester tests cannot run on the client thread, since they can block waiting for a response from the client. However, with Kotlin's [experimental support for coroutines](https://kotlinlang.org/docs/reference/coroutines.html), it's possible to solve this problem.

When writing a test in Kotlin, the test method is written as a `suspend fun`, and annotated with `@CoroutineTest`. Instead of calling blocking methods on the `Client` interface, it makes use of special `suspend` methods. These methods perform the same actions, but suspend the test function instead of blocking.

If you compare [CreeperTest](https://github.com/Aaron1011/Sponge/blob/mctester/src/test/java/org.spongepowered.mod/test/integration/CreeperTest.java) and [CreeperTestKotlin](https://github.com/Aaron1011/Sponge/blob/mctester/src/test/kotlin/org/spongepowered/mod/CreeperTestKotlin.kt) side-by-side, you'll notice that `CreeperTestKotlin` is free of much of the boilerplate in `CreeperTest`. More significantly, it's impossible to cause a thread safety issue by forgetting to wrap operations in `runOnMainThread`. This prevents users from accidentally writing unreliable tests that depend the particular timing of the JUnit and server threads.

Though I've added Kotlin as a dependency of SpongeForge, it's only a test-time dependency. This doesn't commit us to writing any of our implementation in Kotlin, nor does it make the Kotlin libraries available to plugins in the production jar.

**Github bot**

Since McTester runs an actual Minecraft client, we're able to take screenshots. Thanks to XVFB and Mesa, we can even do this on a headless CI server. By default, McTester will automatically take screenshots of failing tests - though this can be configured on a per-test basis via `@ScreenshotOptions`

I've written [a small Github bot](https://github.com/Aaron1011/McTester/tree/master/uploader) that uploads the screenshots to Imgur, and adds a link to the album as a commit status. You can see what it looks like at https://github.com/AaronBot1011/GHTest/pull/2

The uploader is split into two parts - a client that uploads the images, and a server that updates the Github status. This would allow the bot to be enabled for all PRs, not just ones by trusted users.

If there's interest in deploying this bot, I can make the necessary changes to fully integrate it with Travis CI and the SpongeForge repository.

**Future modifications**

There's still room to expand on my initial version of McTester. In particular, it would be fairly straightforward to test against specific versions of certain mods. This could be as simple as ensuring that a particular event continues to work with a mod installed, or as complex as actually using a mod item/block.


build.gradle
```diff
@@ -21,16 +23,29 @@ buildscript {
...
     }
 }
 
+plugins {
+    id 'org.jetbrains.kotlin.jvm' version '1.2.51'
+}
+
+
+repositories {
+    maven { url 'https://jitpack.io' }
```
kashike: Should be deployed to a proper repository before being merged.

kashike (changes requested):

Aaron1011:
One other thing I forgot to mention:

If people agree with the basic concept of this PR, it might make sense to make [McTester](https://github.com/Aaron1011/McTester) part of the SpongePowered organization, similar to [LaunchWrapperTestSuite](https://github.com/SpongePowered/LaunchWrapperTestSuite)

build.gradle
liach: Or aaron cannot afford a maven repo

build.gradle
```diff
@@ -13,6 +13,8 @@ buildscript {
             name = 'forge'
             url = 'http://files.minecraftforge.net/maven'
         }
+        maven { url 'https://jitpack.io' }
+        mavenLocal()
```
liach: This line exists now for debugging purposes, right? Should remove it before official merge.

src/test/java/org.spongepowered.mod/test/integration/ChatTest.java
```diff
@@ -0,0 +1,75 @@
...
+        testUtils.listenOneShot(() -> client.sendMessage(message),
+                new StandaloneEventListener<>(MessageChannelEvent.Chat.class, (MessageChannelEvent.Chat event) -> {
+                    Assert.assertEquals(message, event.getRawMessage().toPlain());
+                }));
+    }
+
+    @Test(expected = AssertionError.class)
+    public void deliberateFailure() throws Throwable {
+        testUtils.listenOneShot(() -> { client.sendMessage("blah"); }, new StandaloneEventListener<>(MessageChannelEvent.Chat.class, (MessageChannelEvent.Chat event) -> {
+            Assert.assertEquals(1, 2);
```
liach: Are you just making sure the event is received after the message's sent? At first glance the purpose of this test does not appear very clear to me.

build.gradle
```diff
@@ -21,16 +23,29 @@ buildscript {
         classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.4'
         classpath 'gradle.plugin.org.spongepowered:spongegradle:0.8.1'
         classpath 'org.spongepowered:mixingradle:0.4-SNAPSHOT'
+        classpath mcTesterDep
```
liach, line 24: Hmm, Would you mind pointing out where `mcTesterDep` is defined? I failed to see that.

build.gradle
```diff
@@ -144,3 +159,27 @@ uploadArchives {
...
+
+//repositories {
+//    mavenLocal()
+//}
+
+dependencies {
+    testCompile mcTesterDep
+    testCompile "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
+    testCompile "org.jetbrains.kotlinx:kotlinx-coroutines-core:0.23.4"
+    compile "org.jetbrains.kotlin:kotlin-reflect:1.2.51"
```
liach: Just curious, why isn't this `testCompile`?

src/test/java/org.spongepowered.mod/test/integration/CreeperTest.java
```diff
@@ -0,0 +1,136 @@
...
+            player.setItemInHand(HandTypes.MAIN_HAND, ItemStack.builder().itemType(ItemTypes.SPAWN_EGG).quantity(1).add(Keys.SPAWNABLE_ENTITY_TYPE, EntityTypes.CREEPER).build());
+        });
+
+        testUtils.waitForInventoryPropagation();
+
+        // Look at the ground two blocks in the z direction
+        Vector3d targetPos = this.testUtils.runOnMainThread(() -> CreeperTest.this.testUtils.getThePlayer().getLocation().getPosition().add(0, -1, 2));
+        client.lookAt(targetPos);
+
+        final Creeper[] creeper = new Creeper[1];
```
liach: Eh, this way of capturing a creeper instance looks weird (though I know you cannot use non-final local variables in lambda expressions. Any way to make it look better?)

src/test/java/org.spongepowered.mod/test/integration/InventoryTest.java
```diff
@@ -0,0 +1,57 @@
...
+        });
+
+
+        // We sleep two ticks to guarantee that the client has been updated.
+        // During the next tick, the server will send our inventory changes to the client.
+        // However, we don't want to rely on this happening at any particular point during the tick,
+        // so we wait two ticks to guarantee that the update packets have been sent by the time
+        // our code runs.
+        //testUtils.sleepTicks(2);
+        testUtils.waitForInventoryPropagation();
```
liach: Does this method replace the previous sleep tick method?

src/test/java/org.spongepowered.mod/test/integration/regression/CrossDimensionCommandBlock.java
```diff
@@ -0,0 +1,110 @@
...
+
+        assertThat("Failed to open command block gui!", guiClass, Matchers.equalTo(Optional.of("net.minecraft.client.gui.GuiCommandBlock")));
+
+        /*this.testUtils.runOnMainThread(() -> {
+            InventoryArchetype type = McTester.getThePlayer().getOpenInventory().orElseThrow(() -> new AssertionError("Failed to open command block gui!")).getArchetype();
+
+            assertThat(type, Matchers.equalTo(InventoryArchetypes.)
+
+            Assert.assertThat(McTester.getThePlayer().getOpenInventory().orElse(null), Matcher);
+        });*/
```
liach: Yes, command block gui does not have an inventory. Can we just use forge code here to check the gui type open? `Minecraft.getMinecraft.openGui`

liach:
@Aaron1011 One of the biggest issue is that if the tests can be run on main thread, we can avoid a lot of unnecessary synchronization, reduce lambdas and ugly array code, and possibly enjoy a speedup. Can you do that?

src/test/java/org.spongepowered.mod/test/integration/ChatTest.java
Aaron1011: I'll add some more comments to the test, and expand on the McTester javadocs. Essentially, `client.sendMessage` only returns once the server has received and processed the chat packet from the client.

build.gradle
Aaron1011: `gradle.properties`

src/test/java/org.spongepowered.mod/test/integration/CreeperTest.java
Aaron1011: I just went with Intellij's suggestion - I don't know of a nicer way.

src/test/java/org.spongepowered.mod/test/integration/regression/CrossDimensionCommandBlock.java
Aaron1011: I've deliberately avoided touching `Minecraft` in any of the test classes. See https://github.com/Aaron1011/McTester/issues/2 for an explanation.

Aaron1011:
@liach: See the 'Kotlin support' of the PR description for more information about running tests on the main thread.

It's fundamentally impossible to run Java-based McTester tests on the server thread. Any `Client` methods, such as `Client.sendMessage`, block until the client has send a response. If the test is running on the main thread, the server will have no way of processing the packet, since the main thread will be stuck in the test class.

The reason that Kotlin-based tests like `CreeperTestKotllin` are able to run on the main thread is because of coroutines. With coroutines, McTester is able to suspend the test method when a blocking method is called, allowing the server thread to make progress and process packets. When the response from the client comes in, the test method is resumed on the main thread.

Unless Java were to add support for coroutines (which I seriously doubt will ever happen), Kotlin-based tests are the only way to run tests on the server thread.

src/test/java/org.spongepowered.mod/test/integration/InventoryTest.java
Aaron1011: Yes - I've removed the comment in a new commit.

Aaron1011:
Oddly, the last two builds failed on Oracle java only, for two unrelated reasons. It's possible that there's a race condition somwhere - I'll investigate more tomorrow.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -1,6 +1,6 @@
-#Sun Jan 08 11:40:19 CET 2017
+#Mon May 07 17:43:54 EDT 2018
 distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-bin.zip
+distributionUrl=https\://services.gradle.org/distributions/gradle-3.3-all.zip
```
ImMorpheus: Why ?

Aaron1011:
After running the tests several times locally, I was able to reproduce the hang twice. Both times, it occured when using Oracle JDK and XVFB.

ghost:
Copied the wrong link 😅

Aaron1011:
With all of the race conditions fixed, and the unnecessary debug messages removed from McTester, this PR is ready for final review.

build.gradle
```diff
@@ -21,16 +21,28 @@ buildscript {
...
     }
 }
 
+plugins {
+    id 'org.jetbrains.kotlin.jvm' version '1.2.51'
+}
+
+
+repositories {
+    mavenLocal()
```
stephan-gh: Is this still needed?

build.gradle
```diff
@@ -144,3 +156,27 @@ uploadArchives {
 }
 
 apply from: 'changelog.gradle'
+
+//repositories {
+//    mavenLocal()
```
stephan-gh: Exists above already

src/main/java/org/spongepowered/mod/mixin/core/fml/common/MixinFMLCommonHandler.java
```diff
@@ -0,0 +1,50 @@
+package org.spongepowered.mod.mixin.core.fml.common;
```
stephan-gh: License header

.travis.yml
```diff
@@ -1,17 +1,38 @@
...
+  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
+  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
+cache:
+  directories:
+    - $HOME/.gradle/caches/
+    - $HOME/.gradle/wrapper/
+    - $TRAVIS_BUILD_DIR/.gradle/
+    - $TRAVIS_BUILD_DIR/build/
+
+
```
liach: extra blank line

build.gradle
```diff
@@ -21,16 +21,23 @@ buildscript {
         classpath 'com.github.jengelman.gradle.plugins:shadow:1.2.4'
         classpath 'gradle.plugin.org.spongepowered:spongegradle:0.8.1'
         classpath 'org.spongepowered:mixingradle:0.4-SNAPSHOT'
+        classpath mcTesterDep
     }
 }
 
+plugins {
+    id 'org.jetbrains.kotlin.jvm' version '1.2.51'
```
liach: Will this allow kotlin code in main source set, too? If yes we might want to avoid that.

src/test/java/org.spongepowered.mod/test/integration/CreeperTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+            assertThat(event.getEntities(), hasSize(1));
+            creeper[0] = (Creeper) event.getEntities().get(0);
+
+            assertTrue("Cause doesn't contain player: " + event.getCause(), event.getCause().contains(CreeperTest.this.testUtils.getThePlayer()));
+            assertTrue("Cause doesn't contain correct item: " + event.getCause(),
+                    event.getCause().getContext().get(EventContextKeys.USED_ITEM).map(i -> i.getType().equals(ItemTypes.SPAWN_EGG)).orElse(false)
+            );
+
+            Sponge.getEventManager().unregisterListeners(this);
```
liach: Is the game instance recycled after each test? (not world instance) If yes you don't need to unregister the listener.

src/test/java/org.spongepowered.mod/test/integration/CreeperTest.java
```diff
@@ -0,0 +1,161 @@
...
+
+        }));
+        client.rightClick();
+
+        assertThat("Creeper did not spawn!", creeper[0], instanceOf(Creeper.class));
+
+
+        EventListener<MoveEntityEvent> moveListener = this.testUtils.listen(new StandaloneEventListener<>(MoveEntityEvent.class, (MoveEntityEvent event) -> {
+
+            if (event.getTargetEntity().getUniqueId() == creeper[0].getUniqueId()) {
```
liach: Use `equals` instead.

src/test/java/org.spongepowered.mod/test/integration/CreeperTest.java
```diff
@@ -0,0 +1,161 @@
...
+            try {
+                this.client.rightClick();
+            } catch (Throwable e) {
+                throw new RuntimeException(e);
+            }
+        }, new StandaloneEventListener<>(PrimeExplosiveEvent.Pre.class, (PrimeExplosiveEvent.Pre primeEvent) -> {
+            assertThat(primeEvent.getTargetEntity(), equalTo(creeper[0]));
+            fuseDuration[0] = primeEvent.getTargetEntity().get(Keys.FUSE_DURATION).get();
+
+
```
liach: useless blank lines

src/test/java/org.spongepowered.mod/test/integration/RegressionTest.java
```diff
@@ -0,0 +1,31 @@
...
+ * IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+ * FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+ * AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+ * LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+ * OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+ * THE SOFTWARE.
+ */
+package org.spongepowered.mod.test.integration;
+
+public @interface RegressionTest {
```
liach: Hmm, can you add retention policy and target for this annotation at least? You can add some javadocs, too.

src/test/java/org.spongepowered.mod/test/integration/UseItemStackTest.java
```diff
@@ -0,0 +1,288 @@
...
+    @Test
+    public void testNormalCycle() throws Throwable {
+        ItemStack stack = this.setupItemStack();
+
+        final UseItemStateHolder useItemStateHolder = new UseItemStateHolder();
+        final int[] firstDuration = new int[1];
+
+        this.testUtils.listen(new StandaloneEventListener<>(UseItemStackEvent.class, new EventListener<UseItemStackEvent>() {
+
+
```
liach:
```java
(UseItemStackEvent event) -> {
    useItemStateHolder.transition(event);
    if (event instanceof UseItemStackEvent.Start) {
        firstDuration[0] = event.getOriginalRemainingDuration();
    }
}
```
for the second parameter for standaloneeventlistener constructor.

src/test/java/org.spongepowered.mod/test/integration/regression/BreakBlockUsedItem.java
```diff
@@ -0,0 +1,88 @@
...
+            gotPlayerEvent[0] = true;
+            Assert.assertThat("Cause doesn't contain EventConextKeys.USED_ITEM", event.getCause().getContext().get(EventContextKeys.USED_ITEM),
+                    Matchers.equalTo(Optional.of(stack.createSnapshot())));
+
+        }));
+
+        Assert.assertTrue("ChangeBlockEvent.Break was not fired!", gotPlayerEvent[0]);
+
+
+
```
liach: Excessive blank lines

src/test/java/org.spongepowered.mod/test/integration/regression/CrossDimensionCommandBlock.java
```diff
@@ -0,0 +1,111 @@
...
+                            .generator(GeneratorTypes.FLAT)
+                            .build("mctester:command_block_world", "Command block World"));
+
+            World world = Sponge.getServer().loadWorld(worldProperties).get();
+
+            this.testUtils.getThePlayer().setLocation(world.getSpawnLocation());
+            this.testUtils.getThePlayer().setItemInHand(HandTypes.MAIN_HAND, ItemStack.of(ItemTypes.COMMAND_BLOCK, 1));
+
+            Location<World> target = this.testUtils.getThePlayer().getLocation().add(0, 1, 1);
+            return target;
```
liach: just return the location without storing it in a local variable

src/test/kotlin/org/spongepowered/mod/FakeEventUser.java
```diff
@@ -0,0 +1,20 @@
+package org.spongepowered.mod;
+
+import org.spongepowered.api.event.Event;
+import org.spongepowered.api.event.EventListener;
+
+public class FakeEventUser<T extends Event> implements EventListener<T> {
```
liach: Why is this java file in a kotlin package?

src/test/java/org.spongepowered.mod/test/integration/InventoryTest.java
```diff
@@ -0,0 +1,56 @@
+package org.spongepowered.mod.test.integration;
```
ImMorpheus: License

src/test/kotlin/org/spongepowered/mod/CreeperTestKotlin.kt
```diff
@@ -0,0 +1,137 @@
+package org.spongepowered.mod
```
ImMorpheus: License

src/test/kotlin/org/spongepowered/mod/EventListenerBug.kt
```diff
@@ -0,0 +1,17 @@
+package org.spongepowered.mod
```
ImMorpheus: License

src/test/kotlin/org/spongepowered/mod/FakeEventUser.java
```diff
@@ -0,0 +1,20 @@
+package org.spongepowered.mod;
```
ImMorpheus: License

liach:
For license issues pointed out, aaron should add a line like this https://github.com/SpongePowered/SpongeAPI/blob/a1418ae3fccbfa1fbc56e05d2f3375dcd1344e1c/gradle/sponge.gradle#L119

```gradle
license.include '**/*.kt'
```

src/test/java/org.spongepowered.mod/test/integration/InventoryTest.java
liach: The file path is wrong too. It needs to be corrected.

build.gradle
```diff
@@ -144,3 +151,23 @@ uploadArchives {
...
+    testCompile "org.jetbrains.kotlinx:kotlinx-coroutines-core:0.23.4"
+    testCompile "org.jetbrains.kotlin:kotlin-reflect:1.2.51"
+}
+
+kotlin {
+    experimental {
+        coroutines 'enable'
+    }
+}
+
```
liach, line 183: Add a line `license.include **/*.kt` to auto format kotlin files for their license.

src/test/java/org.spongepowered.mod/test/integration/UseItemStackTest.java
```diff
@@ -0,0 +1,288 @@
...
+        this.client.holdRightClick(false);
+
+        int expectedDuration = ((Item) stack.getType()).getMaxItemUseDuration((net.minecraft.item.ItemStack) (Object) stack);
+        assertThat("Unexpected item use duration!", firstDuration[0], is(expectedDuration));
+
+        List<UseItemState> expectedStates = Lists.newArrayList(UseItemState.START);
+        for (int i = 0; i < firstDuration[0]; i++) {
+            expectedStates.add(UseItemState.TICK);
+        }
+        expectedStates.addAll(Lists.newArrayList(UseItemState.FINISH, UseItemState.REPLACE, UseItemState.RESET));
```
ImMorpheus: `Collections#addAll`

src/test/java/org.spongepowered.mod/test/integration/UseItemStackTest.java
```diff
@@ -0,0 +1,288 @@
...
+                    firstDuration[0] = event.getOriginalRemainingDuration();
+                }
+            }
+        }));
+
+        this.client.holdRightClick(true);
+        this.testUtils.sleepTicks(40);
+        this.client.holdRightClick(false);
+
+        int expectedDuration = ((Item) stack.getType()).getMaxItemUseDuration((net.minecraft.item.ItemStack) (Object) stack);
```
ImMorpheus: `ItemStackUtil#toNative`

build.gradle
Aaron1011: I don't believe so - all of the actual Kotlin library dependencies (e.g. `kotlin-stdlib`) are set to `testCompile`.

src/test/java/org.spongepowered.mod/test/integration/CreeperTest.java
Aaron1011: No - one Minecraft client is used throughout all of the tests

src/test/kotlin/org/spongepowered/mod/FakeEventUser.java
Aaron1011: I added this when debugging a strange Intellij issue, which appears to have mysteriously gone away. I'll remove it.

src/test/kotlin/org/spongepowered/mod/CreeperTestKotlin.kt
Aaron1011: Fixed

src/test/java/org.spongepowered.mod/test/integration/InventoryTest.java
Aaron1011: Fixed

build.gradle
Aaron1011: Fixed

Aaron1011:
For some reason, gradle thinks that `setupDecompWorkspace` is out-of-date when run on Travis CI, even though I'm caching all of the relevant directories.

Can someone who's more familiar with Gradle comment on why this might be happening? It's unnecessarily increasing build times.

src/test/java/org.spongepowered.mod/test/integration/RegressionTest.java
liach: Fixed

src/main/java/org/spongepowered/mod/mixin/core/client/server/MixinIntegratedServer.java
```diff
@@ -70,10 +70,13 @@ public void shutdown() {
         }
 
         this.mc.addScheduledTask(() -> {
-            // Need to null check in-case this is invoked very early in login
-            if (this.mc.world != null) {
+            // Vanilla calls this, but it's completely unecessary.
```
gabizou, line 73: Why would it be unnecessary? Wouldn't this cause issues in forge clients where this could be causing client threads to hang?

src/test/java/org/spongepowered/mod/test/integration/CreeperTest.java
```diff
@@ -0,0 +1,159 @@
...
+public class CreeperTest extends BaseTest {
+
+    public CreeperTest(TestUtils testUtils) {
+        super(testUtils);
+    }
+
+    @Test
+    public void explodeCreeper() throws Throwable {
+        this.testUtils.runOnMainThread(() -> {
+            Player player = this.testUtils.getThePlayer();
```
gabizou: This is SpongeAPI dependent?

gabizou: Is McTester only for Sponge API testing?

src/test/java/org/spongepowered/mod/test/integration/RegressionTest.java
```diff
@@ -0,0 +1,50 @@
...
+import java.lang.annotation.Target;
+
+/**
+ * An annotation used to mark tests that ensure that a specific
+ * Sponge issue has not regressed. These will usually, but not always,
+ * be written using Mctester.
+ */
+@Retention(RetentionPolicy.RUNTIME)
+@Target(ElementType.TYPE)
+public @interface RegressionTest {
```
gabizou: How does this regression test test sponge issues? Is it that we provide these tests like the `testplugins` listeners to verify that they work as expected?

gabizou (changes requested):
I like the documentation on test plugins, but I'm just failing to see how the regrssion tests will work unless we add regression tests for each major issue closed.

src/main/java/org/spongepowered/mod/mixin/core/client/server/MixinIntegratedServer.java
Aaron1011: The `loadWorld(null)` call already shuts down the server and kicks the client. See the rest of the comment in the file for why it can cause issues.

src/test/java/org/spongepowered/mod/test/integration/CreeperTest.java
Aaron1011: At the moment, yes. Its possible that I'll add support for testing pure Forge mods at some point in the future.

src/test/java/org/spongepowered/mod/test/integration/RegressionTest.java
Aaron1011: Yes, but regression tests also direct the client to actually trigger the listeners. For example, `CrossDimensionCommandBlock` makes the client right-clock a command block after changing dimensions, and verifies that the proper GUI actually opens client-side.

Aaron1011:
@gabizou: I think there's value to adding regression tests, even if we don't cover every single issue.

For example, `CrossDimensionCommandBlock` tests an issue that has already regressed once after the initial fix. Having this test in place means that we don't need to remember to periodically check it, nor wait around for a user to report it should it occur again

ryantheleach:
Other tests I would do just off the top of my head.

End Dimension "respawning hacks"

Go to the end, make sure the platform exists. 
Kill the dragon (using commands or whatever)
Teleport the player back to the overworld.
Make sure client can interact with things in the overworld.
Maybe some sanity tests about the player being copied / inventory?

World Border issues with game crashing.

Gravestone mod regression testing.

and Maybe something to do with our isOP tests?

**Edit:** Apologies if it sounded like I wanted this added to the PR. The above was intended to be notes / future work, as a +1 I like this, we need this, here is some future possible tests that could build off of this work.

Aaron1011:
@ryantheleach: Those all sound like good ideas, but they would make more sense in a separate PR. This PR is already pretty large as it is.

Aaron1011:
This is currently blocked on getting the necessary X11 libraries installed on Forge Jenkins.

@progwml6: What do you think the best away is to go about doing this?

Aaron1011:
I've managed to create [a Docker image](https://github.com/Aaron1011/mc-headless) containing all of the necessary X11 dependencies. SpongeForge can now successfully run McTester on Jenkins via the [mctesternew](https://github.com/SpongePowered/SpongeForge/tree/mctesternew) branch.

If all goes well, I plan to merge this PR sometime this week. I may put the McTester tests in a separate gradle task, to make it easier for people to skip running them locally.

build.gradle
```diff
@@ -144,3 +151,26 @@ uploadArchives {
...
+    testCompile mcTesterDep
+    testCompile "org.jetbrains.kotlin:kotlin-stdlib-jdk8"
+    testCompile "org.jetbrains.kotlinx:kotlinx-coroutines-core:0.23.4"
+    testCompile "org.jetbrains.kotlin:kotlin-reflect:1.2.51"
+    testCompile "org.apache.logging.log4j:log4j-core:2.8.1"
+}
+
+kotlin {
+    experimental {
+        coroutines 'enable'
```
stephan-gh: Can you update this/McTester to Kotlin 1.3? As far as I know, coroutines are declared stable in 1.3, so we'd be able to remove this.

.travis.yml
```diff
@@ -1,17 +1,37 @@
...
 git:
   submodules: false
-before_install: scripts/travis-initialize-submodules
+before_install:
+  - scripts/run-xvfb
+  - scripts/travis-initialize-submodules
 
 install: ./gradlew setupDecompWorkspace
-script: ./gradlew build
+script: DISPLAY=:99 ./gradlew test --stacktrace
```
stephan-gh: We still need to run the other tests (e.g. license header checks) as well. Is there a reason you have changed this to run `test` instead of `build`?

stephan-gh (changes requested):

.travis.yml
Aaron1011: That was unintentional - I'll change it back.

build.gradle
Aaron1011: Fixed

ryantheleach:
@Aaron1011 Due to the size of this PR, can you add a checklist of remaining tasks / what you need from others into the top post?

Aaron1011:
@ryantheleach: I'm currently trying to resolve a weird Gradle issue. Once that's fixed, this should be ready to merge.

build.gradle
gabizou: Could the dependency be normalized so it's not classpath based? As we will eventually be moving to Kotlin-dsl, classpath provided plugins are annooying since they have to be applied in the old deprecated methods, versus applying via the `plugins {}` format.

More can be read here: https://docs.gradle.org/current/userguide/kotlin_dsl.html#sec:plugins_resolution_strategy

.travis.yml
```diff
@@ -1,17 +1,37 @@
...
+
+before_cache:
+  - rm -f  $HOME/.gradle/caches/modules-2/modules-2.lock
+  - rm -fr $HOME/.gradle/caches/*/plugin-resolution/
+cache:
+  directories:
+    - $HOME/.gradle/caches/
+    - $HOME/.gradle/wrapper/
+    - $TRAVIS_BUILD_DIR/.gradle/
+    - $TRAVIS_BUILD_DIR/build/
```
stephan-gh: Why do you cache the build files? This directory should be empty on each Travis build to ensure a clean re-build.

Additionally, files in there will definitely change with every run, and re-uploading is slow. With a bit of luck, builds will be as slow as without the cache. (Which is the main reason why we haven't been using the cache yet...)

Jenkinsfile
```diff
@@ -23,7 +28,7 @@ pipeline {
                         archiveArtifacts artifacts: 'build/changelog.txt', fingerprint: false
                 }
             }
-        }
+        }/*
```
stephan-gh: This should be reverted before merging this PR.

stephan-gh (changes requested):

stephan-gh (approved):
(I have only checked the build system changes)

Issue #2414, opened 2018-08-28
ChannelRegistrationEvent never contains player in cause


mmis1000:
**I am currently running**

- SpongeForge version: 3422
- Forge version: OpenJDK 8
- Java version: linux
- Operating System: Galaxy, LuckPerms, SpongePrometheusExporter, VillageMarkerSponge, AppleSkin, Duplicate Entity UUID Fix (DEUF), JMFixer, JourneyMap, Just Enough Items, RandomPatches

**Issue Description**

Sponge didn't add player to cause out of main thread because it may be unsafe,  
but ChannelRegistrationEvent always happened on netty thread.  
So, alway no player in cause.  

Detail explanation in the commit by Zidane 
https://github.com/SpongePowered/SpongeForge/pull/2415
Just make an issue in case of this issue may be forgotten.

PR #2415, opened 2018-08-28
Provide player to ChannelRegistrationEvent.

Zidane:
Fixes: https://github.com/SpongePowered/SpongeForge/issues/2414

This event fires on the Netty thread in Forge and we cannot delay a tick to fire it on the main thread (as a plugin dev may want to quickly unregister a channel).

Remove pointless nethandler context as we cannot make a frame here. Instead, provide the Platform type being registered on. I would appreciate feedback here.

Signed-off-by: Chris Sanders <zidane@spongepowered.org>

### DIFF
--- a/src/main/java/org/spongepowered/mod/network/SpongeModNetworkManager.java
+++ b/src/main/java/org/spongepowered/mod/network/SpongeModNetworkManager.java
@@ -35,7 +35,7 @@ import com.google.common.collect.Maps;
 import com.google.inject.Singleton;
 import net.minecraft.client.Minecraft;
 import net.minecraft.client.entity.EntityPlayerSP;
-import net.minecraft.network.INetHandler;
+import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.network.NetHandlerPlayServer;
 import net.minecraft.network.PacketBuffer;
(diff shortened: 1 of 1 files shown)


src/main/java/org/spongepowered/mod/network/SpongeModNetworkManager.java
```diff
@@ -68,54 +66,35 @@
...
-            return "NetHandler";
+        if (event.getSide().isClient()) {
+            player = null; // TODO In 1.13, Client player will be a Sponge player, should we still put in the cause in 1.12?
+        } else {
+            player = ((NetHandlerPlayServer) event.getHandler()).player;
         }
-    };
 
-    private final Map<String, SpongeModChannelBinding> channelMap = Maps.newHashMap();
+        final Cause currentCause = Cause.of(EventContext.empty(), player == null ? Sponge.getGame() : player, event.getSide().isClient() ?
```
gabizou, line 82: Is is not on the main thread? Can't imagine why we wouldn't be able to use the cause stack manager to grab the parenting mod container.

ImMorpheus:
@Zidane What's the status of this ?

Is this ready for review ?

PR #2604, opened 2019-02-11
Fix NPE with SidedDelegate

LXGaming:
Fixes `NullPointerException` caused by calling `getClientToServerNetworkManager`. This is preventing the Server side from starting

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeMod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeMod.java
@@ -154,6 +154,9 @@ public class SpongeMod extends MetaModContainer {
             if (SERVER_THREAD != null) {
                 return SERVER_THREAD == current;
             }
+            if (FMLCommonHandler.instance().getSidedDelegate() == null) {
+                return false;
+            }
             final NetworkManager client = FMLCommonHandler.instance().getClientToServerNetworkManager();
             // Here we're just checking if we're connected to a server and
             // connected to that server, because the connection would no longer


kashike (approved):

kashike:
Thanks, @LXGaming.

liach:
hmm, is there an example crash? i feel this fix is somewhat weird because users should not access server when the game is initializing.

LXGaming:
@liach https://gist.github.com/LXGaming/0683f1e9beaa378b0efffaefb1fd29b3

Bug is present in SpongeForge-1.12.2-2768-7.1.6-RC3564

liach:
After looking at the mixins, I believe your fix is wrong.

https://github.com/SpongePowered/SpongeForge/blob/3032bab6ecbdbc9d8444b46c7b362d5bb67f331d/src/main/java/org/spongepowered/mod/mixin/optimization/threadchecks/MixinSpongeImplHooks_ThreadChecks.java#L54

Looking at https://gist.github.com/LXGaming/0683f1e9beaa378b0efffaefb1fd29b3#file-gistfile1-txt-L18, the only chance the erroneous method is called is when `SpongeMod.isClient()` returns true.

In that method, looking at 
https://github.com/SpongePowered/SpongeForge/blob/3032bab6ecbdbc9d8444b46c7b362d5bb67f331d/src/main/java/org/spongepowered/mod/SpongeMod.java#L172-L173
I realize that the fields might had some desync between threads so both fields may be `null`.

As a result, my suggested fix is to add `volatile` on these two fields below instead. They are definitely accessed by multiple threads. 
https://github.com/SpongePowered/SpongeForge/blob/3032bab6ecbdbc9d8444b46c7b362d5bb67f331d/src/main/java/org/spongepowered/mod/SpongeMod.java#L142-L143
The currently merged changes would lead to more side effects:
https://github.com/SpongePowered/SpongeForge/blob/cf2cb731197c950285ecc32a5e3db5a03b6030e7/src/main/java/org/spongepowered/mod/mixin/optimization/threadchecks/MixinSpongeImplHooks_ThreadChecks.java#L49-L57
With the current change, the `isMainThread` method would return `false` now, breaking the game even more. @kashike I sincerely suggest you to revert this merge.

LXGaming:
Ah yep, I saw the `isClient` check but didn't think twice about it.

I don't believe setting it to volatile will fix it as the issue is due to side being set to CLIENT as a fallback

Upon a quick test `isMainThread` does appear to function correctly on Client and Server sides due to the checks inside of `isClientRunningServerAndServerThread` (Which explains why nothing exploded when I tested the fix initially), The current issue I can see is that `isMainThread` now returns false if the server thread is null when previously it returned true.

liach:
Setting `volatile` can fix because when the static field is updated in one thread, the other thread might not see the update and still see it as `null`, leading to a situation that both threads are `null`.

liach:
Moreover, can `SpongeMod.isClient` be changed to check if `net.minecraft.server.DedicatedServer` class exists? That method is not actually thread aware.

Dockter:
This is NOT fixed.  Issue remains.

PR #2611, opened 2019-02-15
Fix for FG 2.3 changes stable-5

clienthax:


ImMorpheus:
Quick disclaimer:
This is an unsupported version (1.10) of SpongeForge. Support for 1.10 ended in January 2018, which means no support from the team is expected if something breaks.

PR #2613, opened 2019-02-15
Fix for FG changes - stable-7

clienthax:


gabizou:
Yeah, no... this broke version strings for builds and deploying to maven. Fixing, but please, double check artifact names/versions after the fact before merging!

Lignium:
SpongeVanilla also broke. Fix someone please, it's impossible to compile...

clienthax:
@Lignium make an issue against SV

PR #2791, opened 2019-06-02
Add Integration Tests for Schematics

gabizou:
[SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2278) | **SpongeForge**

Adds MCTester based integration tests for placing schematics and verifying the placed schematic will output the same exact schematic.

Note that due to the nature of world ticking, we cannot test against `TileEntity` instances that change their data per tick (burning furnaces, etc).

Note that this does NOT test entities actively as entities datas change far too dynamically, and would require manual (read hardcoded) pruning of dynamic data.


src/integration-test/java/org/spongepowered/mod/test/integration/regression/schematic/PlacingSchematicTest.java
```diff
@@ -0,0 +1,161 @@
...
+              and then "translating" and "re-translating back" into a Schematic will yeild the an equal Schematic.
+            - The full circle of creating, serializing, deserializing, and translating produces the same
+              schematic as if you loaded a new schematic from file.
+            - Schematics that contain extra data attached to tile entities are retained, but the engine
+              may not respect those when a new schematic is created from the placed original schematic
+             */
+            assertEquals(toPlace, deserialized);
+            return schematica;
+        });
+        this.testUtils.waitForWorldChunks();
```
gabizou, line 146: @Aaron1011 this is where I'm kinda struggling to figure out what to do next for future integration tests. I want to be able to see where the schematic ends up, and maybe start building a "redstone" test to trace the `ChangeBlockEvent`s etc.

src/integration-test/java/org/spongepowered/mod/test/integration/regression/schematic/PlacingSchematicTest.java
```diff
@@ -0,0 +1,161 @@
...
+    // The origin point where to apply all schematics.
+    public static final Vector3i ORIGIN = new Vector3i(0, 64, 0);
+    public static final Vector3i OFFSET = new Vector3i(4, 2, 4);
+
+
+    public PlacingSchematicTest(TestUtils testUtils) {
+        super(testUtils);
+    }
+
+    @BeforeClass
```
gabizou, line 80: @Aaron1011 btw,  this does work quite well, runs before the class is initialized, and the tile entities are being registered with forge.

PR #2970, opened 2019-09-09
Upgrade Buildscript for the New SpongeGradle

gabizou:
[SpongeGradle](https://github.com/SpongePowered/SpongeGradle/pull/22) | [SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/1980) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2374) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/419) | **SpongeForge**

This is a **working** PR with the development lock step with **SpongeGradle**. While this technically is compatible with Gradle 4.10.3, I've elected to use the latest Gradle 5.6.2 (as of current writing) to take advantage of the improved Kotlin-DSL since those versions, and configuration avoidance throughout the build process. 

Some changes:
- Constants are just constants in files
- Things that should've been made properties are now properties and/or functions in SpongeGradle or in the `gradle.properties` of their respective projects
- Updated to use appropriate Gradle ideologies with configuration avoidance (lazy evaluation)
- Reduced the overall build file size and using as much handling with the gradle plugins to configure most things we've otherwise duplicated and/or scripted with other scripts.


buildSrc/build.gradle.kts
```diff
@@ -0,0 +1,19 @@
...
+    `kotlin-dsl`
+    `java-library`
+}
+
+repositories {
+    mavenLocal()
+    mavenCentral()
+    gradlePluginPortal()
+    jcenter()
+    maven(url = "https://files.minecraftforge.net/maven")
```
JBYoshi, line 11: This will also need the Sponge Maven repository once merged, although I don't know where in the order it ought to go (if it's above `mavenLocal()`, published versions would take priority; if it's below, unpublished versions would take priority). Same thing with SAPI, SC, and SV.

JBYoshi (changes requested):

PR #3131, opened 2020-04-15
Fix missing ShouldFires 

ImMorpheus:


ImMorpheus:
This can now be merged without the SC PR.

It should fix a few issues with plugins (ShouldFire flag used for the wrong event) and stop firing unneeded events.

PR #3149, opened 2020-05-02
Map api

tyhdefu:
[SpongeAPI](https://github.com/SpongePowered/SpongeAPI/pull/2112) | [SpongeCommon](https://github.com/SpongePowered/SpongeCommon/pull/2573) | [SpongeVanilla](https://github.com/SpongePowered/SpongeVanilla/pull/429) | SpongeForge

Literally just to implement MapDataBridge, because forge changes it to an int, and seperate implementations are required.

### DIFF
--- a/src/main/java/org/spongepowered/mod/mixin/api/common/map/MapDataBridgeForgeMixin_API.java
+++ b/src/main/java/org/spongepowered/mod/mixin/api/common/map/MapDataBridgeForgeMixin_API.java
@@ -1,0 +1,46 @@
+/*
+ * This file is part of Sponge, licensed under the MIT License (MIT).
+ *
+ * Copyright (c) SpongePowered <https://www.spongepowered.org>
+ * Copyright (c) contributors
+ *
+ * Permission is hereby granted, free of charge, to any person obtaining a copy
+ * of this software and associated documentation files (the "Software"), to deal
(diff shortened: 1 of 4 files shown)


src/main/java/org/spongepowered/mod/mixin/api/common/map/MapDataBridgeForgeMixin_API.java
```diff
@@ -0,0 +1,46 @@
...
+ */
+package org.spongepowered.mod.mixin.api.common.map;
+
+import net.minecraft.world.storage.MapData;
+import org.spongepowered.asm.mixin.Mixin;
+import org.spongepowered.asm.mixin.Shadow;
+import org.spongepowered.common.bridge.world.storage.MapDataBridge;
+
+@Mixin(MapData.class)
+public abstract class MapDataBridgeForgeMixin_API implements MapDataBridge {
```
dualspiral: This should be in the `o.s.m.mixin.core` package somewhere, not the `api` package.

Should be named `MapDataMixin_Forge`

tyhdefu: I have moved it into a similar place as SV, `org.spongepowered.mod.mixin.core.map`

Issue #3086, opened 2020-02-18
Can't catch iron golem with entity.getType().equals(EntityTypes.IRON_GOLEM)


electron93:
**I am currently running**

- SpongeForge version: **1.12.2-2838-7.1.10-RC3994**
- Java version: 1.8.0_242/64-bit
- Operating System: Linux

**Issue Description**

```java
    @Listener
    public void onSpawnEntity(final SpawnEntityEvent event) {
        event.getEntities().forEach(entity -> {
            final boolean isGolem = entity.getType().equals(EntityTypes.IRON_GOLEM);
            final boolean isCreeper = entity.getType().equals(EntityTypes.CREEPER);
            final boolean isZombie = entity.getType().equals(EntityTypes.ZOMBIE);

            if (isGolem) {
                this.plugin.getLogger().warn("IRON_GOLEM!!!"); // should be displayed when an iron golem spawns
            }

            if (isCreeper) {
                this.plugin.getLogger().warn("CREEPER!!!"); // works
            }

            if (isZombie) {
                this.plugin.getLogger().warn("ZOMBIE!!!"); // works
            }

            this.plugin.getLogger().warn(entity.getType().getName()
                    + " | " + entity.getType().getId()
                    + " | " + entity.getType().getTranslation().getId()
                    + " | " + entity.getType().getTranslation().get()
                    + " | " + entity.getType().getEntityClass().getName()
            );
        });
    }
```

For example:
```
[Server thread/INFO] [plugin]: e1ectron issued server command: /summon minecraft:creeper
[Server thread/WARN] [plugin]: CREEPER!!!
[Server thread/WARN] [plugin]: creeper | minecraft:creeper | entity.Creeper.name | Creeper | net.minecraft.entity.monster.EntityCreeper
[Server thread/INFO] [minecraft/DedicatedServer]: [e1ectron: Object successfully summoned]
```
```
[Server thread/INFO] [plugin]: e1ectron issued server command: /summon minecraft:zombie
[Server thread/WARN] [plugin]: ZOMBIE!!!
[Server thread/WARN] [plugin]: zombie | minecraft:zombie | entity.Zombie.name | Zombie | net.minecraft.entity.monster.EntityZombie
[Server thread/INFO] [minecraft/DedicatedServer]: [e1ectron: Object successfully summoned]
```

But this isn't working for iron golem (villager_golem):
```
[Server thread/INFO] [plugin]: e1ectron issued server command: /summon minecraft:villager_golem
[...]

PR #3173, opened 2020-06-11
Fix entityName when registering SpongeEntityType

electron93:
After the fix for #3086 (in a4f980b), I decided to check: are the entities from the mods being registered correctly?

The short answer is no. Because using the code from #3086 I got the following output:

```
[plugin]: e1ectron issued server command: /summon ic2:electric_boat
[plugin]: boatelectric | ic2:boatelectric | SpongeEntityType{id=ic2:boatelectric, name=boatelectric, translation=SpongeTranslation{id=entity.generic.name}, modid=ic2, class=ic2.core.item.EntityBoatElectric}
```

But it should be like this:

```
[plugin]: e1ectron issued server command: /summon ic2:electric_boat
[plugin]: electric_boat | ic2:electric_boat | SpongeEntityType{id=ic2:electric_boat, name=electric_boat, translation=SpongeTranslation{id=entity.generic.name}, modid=ic2, class=ic2.core.item.EntityBoatElectric}
```

This PR corrects the name of the entity during registration.

---

I tried to check and make sure that I was doing it right. Therefore, it was useful to look at what was happening at MinecraftForge. Looking at all this, I came to the conclusion that this was relevant at the time of Minecraft(Forge) 1.10:

https://github.com/MinecraftForge/MinecraftForge/blob/e4caf46cb954ea86fe338cd256402cd84f824586/src/main/java/net/minecraftforge/fml/common/registry/EntityRegistry.java#L182-L184

### DIFF
--- a/src/main/java/org/spongepowered/mod/SpongeMod.java
+++ b/src/main/java/org/spongepowered/mod/SpongeMod.java
@@ -46,7 +46,6 @@ import net.minecraftforge.fml.client.FMLFolderResourcePack;
 import net.minecraftforge.fml.common.CertificateHelper;
 import net.minecraftforge.fml.common.FMLCommonHandler;
 import net.minecraftforge.fml.common.LoadController;
-import net.minecraftforge.fml.common.ModContainer;
 import net.minecraftforge.fml.common.ModContainerFactory;
 import net.minecraftforge.fml.common.ModMetadata;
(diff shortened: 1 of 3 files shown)


gabizou (approved):
Great investigation and catch! It's true that thinking about that bit of code was definitely relevant at the time because of the lack of standardization (TileEntities are still suffering from this in a very mild fashion last I recall at least).

Looks good to me!

Issue #3204, opened 2020-08-31
EntityMountEvent being captured by SpongeForge


EliseZeroTwo:
**I am currently running**
- SpongeForge version: 1.12.2-2838-7.3.0 
- Forge version: 1.12.2-14.23.5.2854
- Java version: oracle64-1.8.0.112
- Operating System: Tested on both MacOS 10.16 and some Linux flavour
- Plugins/Mods:
    - Minecraft
    - Minecraft Coder Pack
    - SpongeAPI
    - Sponge

**Issue Description**
SpongeForge captures the event EntityMountEvent if the event is a dismount. The event is accessible correctly inside of SpongeForge where as the event is not called inside of Forge on a dismount, the event is called correctly if it is a mount though. Removing SpongeForge makes the event work in Forge.

PR #3272, opened 2021-10-10
Fire RideEntityEvent back to Forge

### LINKED ISSUE #3133: Issue with Sponge and Lycanite mobs’ rideable creatures
I‘m running
SpongeForge version: 1.12.2-2838-7.1.10 and 1.12.2-2838-7.2.0
Forge version: forge-1.12.2-14.23.5.2838
Java version: 8
lycanitesmobs-1.12.2-2.0.7.0-aberrant14.jar
_spongeforge-1.12.2-2838-7.1.10.jar
Issue Description:
Whenever I press shift to go down while riding any creature of the mod, the mob stops moving and I can’t do anything. This only happens with sponge on the server, and with every version of the mod.

### LINKED ISSUE #3254: Issue EntityMountEvent
**I am currently running**

SpongeForge version: SpongeForge 1.12.2-2838-7.2.3

https://gyazo.com/709cbfc68c3897017c3c834b42173509

**Issue Description from Elise**

for some reason, the event from pixelmon/minecraft for the entity dismounting seems to be firing incorrectly
mounting works but dismounting does not in forge server as it doesn't fire the dismount event for some reason

**What I am working on**

I got in spawn pixelmon bikes.
And I want to keep bicycles in spawn. But some people like to have fun with their bike. and throw it in the ocean. So I asked a developer if there is a possibility that we can make the bikes automatically return to their place.

After a certain time, I found a developer called elise, which is working on developing the plugin, but she found problems while working on the plugin. Which are unfixable on her side. Because it seems it are sponge issues.

**She explained to me**

SpongeForge catches an EntityMountEvent event if the event is a dismount. The event is correctly available on SpongeForge where as the event is not triggered inside Forge when dismounting but triggered correctly if it is mount. Removing SpongeForge makes the event run in Forge as expected.

clienthax:
Fixes #3133

Likely fixes
#3254
#3204

### DIFF
--- a/src/main/java/org/spongepowered/mod/event/SpongeToForgeEventFactory.java
+++ b/src/main/java/org/spongepowered/mod/event/SpongeToForgeEventFactory.java
@@ -47,6 +47,7 @@ import net.minecraft.world.storage.DerivedWorldInfo;
 import net.minecraftforge.common.MinecraftForge;
 import net.minecraftforge.event.ServerChatEvent;
 import net.minecraftforge.event.entity.EntityJoinWorldEvent;
+import net.minecraftforge.event.entity.EntityMountEvent;
 import net.minecraftforge.event.entity.EntityTravelToDimensionEvent;
 import net.minecraftforge.event.entity.item.ItemTossEvent;
 import net.minecraftforge.event.entity.living.LivingDeathEvent;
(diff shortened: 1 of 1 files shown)


clienthax:
Not sure if this should only be targeting the Dismount event, due to the overwrite of entityDismount in EntityMixin
From issues at least it appears that only the dismount one would be needed, but unfamiliar with this code.

PR #3288, opened 2022-02-22
Gradle Incremental compilation

shisheng-1:
[Incremental compilation](https://docs.gradle.org/current/userguide/performance.html#incremental_compilation). Gradle recompile only the classes that were affected by a change. This feature is the default since Gradle 4.10. For an older versions, we can activate it by setting `options.incremental = true`.


zml2008:
begone spambot

